PR #14, opened 2020-11-23
Add option to change locale

weph:
Hey @SilentVoid13 

great plugin! One thing I was missing is the ability to change the locale so that the dates are formatted with (in my case) german week and month names.

I added an option to the settings to let the user choose their locale. The dropdown shows all locales available in moment.js.


SilentVoid13:
Hey @weph, sorry for the late response.
Thanks a lot for this contribution, that's a neat feature !
Merging.

PR #18, opened 2020-11-27
Save settings after each change

weph:
It's me again 😉

Thanks for merging my other PR. After restarting Obsidian I noticed that some of my settings were gone.

I've seen that you're saving the settings in the `onunload` handler but that doesn't seem to work. I've added `saveData` calls to all settings that were missing it. I don't know if this is the way to go, but I've seen it in other plugins.


SilentVoid13:
You're right, that's a mistake.
Thanks again for the contribution !

PR #42, opened 2021-01-14
Added "Replace" in active file command

pyjamads:
The added command allows you to replace any templates in your active file.

This is useful if the file has already been created and you just want to add something to it using the custom templates.

add a Templater template (such as ```{{tp_now}}``` anywhere in an open document, and run the Replace in Open File command, to expand that template.


SilentVoid13:
That's a great idea ! Thanks a lot for the contribution !

PR #49, opened 2021-01-20
Weekday insertion internal Templates

Mocca101:
This is my first public pull request so please tell me if there's something I should do differently!

For my Daily Notes I like to have the Weekday as well, so I wanted to be able to have those added automatically as well.

I added two internal Templates which allow you to get the weekdays in a short "Tue" or long "Tuesday" format. Either of the current date or by the title date.

### DIFF
--- a/src/internal_templates.ts
+++ b/src/internal_templates.ts
@@ -25,6 +25,8 @@ export const internal_templates_map: {[id: string]: Function} = {
     "title_tomorrow": tp_title_tomorrow,
     "title_today": tp_title_today, // DEPRECATED
     "title_yesterday": tp_title_yesterday,
+    "date_weekday": tp_date_weekday,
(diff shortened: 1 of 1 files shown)


kaveet:
This is a great contribution for a first pull request, but have you considered using the custom format argument in the `tp_date` template in your Daily Notes instead of adding another internal template? It sounds like the behavior you're after is supported out of the box.

For example:

```markdown
{{tp_date:f=dddd}}
```

Outputs:

```plaintext
Sunday
```

And:

```markdown
{{tp_date:f=dddd}}
```

Outputs:

```plaintext
Sun
```

Mocca101:
I did not realize I could use it like that! 
Thank you very much for telling me and taking your time to look at this request.
And thank you for your awesome plugin.

Issue #200, opened 2021-05-07
Allow users to provide default template for new empty files


ankushg:
**Is your feature request related to a problem? Please describe.**
The current command to creating a new file from template is nice, but can't handle the situation when you follow an internal link to a nonexistent file. Additionally, it would be nice if every new file just included my default template.

**Describe the solution you'd like**

add `default_included_template_file: String` to `TemplaterSettings`
  - default it to `""`
  - add a basic UI like `script_folder`
  - Question: Do we need to rebind the `update_trigger_file_on_creation` trigger?
  - stretch goals: 
    - hide `default_included_template_file` behind `update_trigger_file_on_creation` setting the same way `templates_pairs` are hidden
    - stretch goal: autocomplete filepath for default template like Periodic Notes does

modify `update_trigger_file_on_creation` logic to 
```js
// Require empty file to avoid clobbering other plugins which specify file content (Calendar, Daily Notes, etc)
if (file.stat.size == 0 && this.settings.default_included_template_file) {
  const includeStatement = `<% tp.file.include("[[${this.settings.default_included_template_file}]]") %>`;
  await this.vault.modify(file, includeStatement);
}
this.templater.overwrite_file_templates(file);
```

**Additional context**
I'd be happy to submit a PR for this functionality if it lines up with your vision for the project.

PR #203, opened 2021-05-07
feat: Allow users to provide a default template for empty new files

ankushg:
Addresses https://github.com/SilentVoid13/Templater/issues/200

- requires `trigger_on_file_creation` to be enabled
- only runs on files with no content

<details>
  <summary>For an example of what a default template might look like:</summary>

```js
<%*
// This logic is actually defined as a Script User Function
// so I can just do:
// const defaultTemplate = tp.user.getDefaultTemplate(tp);
const fallback_template = "[[assets/templates/fallback]]";
const config = [
  { 
    regex: /^\d{4}-\d{2}-\d{2}$/,
    template: "[[assets/templates/daily_note]]"
  },
  { 
    regex: /^\d{4}-\d{2}-W\d{2}$/,
    template: "[[assets/templates/weekly_note]]"
  },
  { 
    regex: /^\d{4}-\d{2}$/,
    template: "[[assets/templates/monthly_note]]"
  },
  {
    regex: /^\d{4}-Q\d{1}$/,
    template: "[[assets/templates/quarterly_note]]"
  },
  {
    regex: /^\d{4}$/,
    template: "[[assets/templates/annual_note]]"
  },
];

const title = tp.file.title;

const matching_default = config
  .find(item => item.regex.test(title));

if (matching_default) {
  const default_template = matching_default.template;
} else {
  const default_template = fallback_template;
}
-%>
<% tp.file.include(default_template) -%>
```
</details>

### DIFF
--- a/src/Settings.ts
+++ b/src/Settings.ts
@@ -102,8 +102,12 @@ export class TemplaterSettingTab extends PluginSettingTab {
 						this.plugin.settings.trigger_on_file_creation = trigger_on_file_creation;
 						this.plugin.saveSettings();
 						this.plugin.update_trigger_file_on_creation();
+						// Force refresh
+						this.display();
 					});
 			});
+
+		if (this.plugin.settings.trigger_on_file_creation) {
 			desc = document.createDocumentFragment();
 			desc.append(
 				"Templater will automatically apply this template file to new empty files as they are created.",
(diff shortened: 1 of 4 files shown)


ankushg:
@SilentVoid13 I'm finding that I'm sometimes hitting the limit on nested includes.

Is this some sort of known side effect of using includes within includes? Not sure how includes work behind-the-scenes but I wouldn't be surprised if they create a file, which triggers this code, which tries to include something in a loop 🙈 

Maybe we need an additional check other than `file.stat.size == 0` to also make sure it's not created as a result of an include?

SilentVoid13:
Hey @ankushg sorry for the late response. Thanks a lot for the contribution! Fixed a few things in your PR, everything should be working now.
Merging, this will be available in next release!

Issue #194, opened 2021-05-05
Documentation dead links


Oak86:
Wasn't sure if this is the right place for it, but I found that several of the links here are broken:
https://silentvoid13.github.io/Templater/docs/internal-variables-functions

In particular, I tried these and none of them worked:

Config module: tp.config
Date module: tp.date
File module: tp.file
Frontmatter module: tp.frontmatter
Obsidian module: tp.obsidian
System module: tp.system
Web module: tp.web

Let me know if there's a better way to report documentation issues!

andrew8088:
Looks like this is actually [a known docusaurus bug](https://github.com/facebook/docusaurus/issues/3372).

SilentVoid13:
Hey @Oak86 @andrew8088, you're right that is a docusaurus bug. I managed to "bypass" it by changing the links format.
Thanks a lot for the report, that helps a lot! Don't hesitate to open an issue when you notice something like this.

PR #207, opened 2021-05-10
Fix broken documentation links

andrew8088:
Fixes the broken documentation links reported in #194


andrew8088:
On close inspection, I'm not sure this actually fixes it ... when I run docusaurus locally, it complains that these links are broken ...

Issue #188, opened 2021-05-04
unable to insert dynamic commands in templates


weeyizhi:
**Plugin informations (please complete the following information):**
 - OS: Windows 10
 - Templater version: 1.6.0
 - Obsidian version: 0.11.13
 - Templater settings: Template folder location: "templater", timeout: 5s

**Describe the bug**
The new dynamic commands are great, but I can't put one into my templates as they get computed into something weird.
Consider the following template:
```
# <%tp.file.title%>

<%+ tp.date.now() %>
```

after applying the template on a file, we get:
```
# 2021-05-04

NaN
```

**Expected behavior**
Dynamic commands should not be computed when applying templates. Or maybe provide the user some way to escape certain blocks.

SilentVoid13:
Hey @weeyizhi you're right, I'll try to fix that in a future update.

SilentVoid13:
Hey @weeyizhi @Moonbase59, this should now be fixed as of Templater 1.7.0. Update and tell me if this works now!

Moonbase59:
@SilentVoid13 Yup, well done, thanks! Works fine now.

perymimon:
hi Im on Templater 1.16 and that bug happen again

PR #214, opened 2021-05-11
Do not replace dynamic templates on initial expansion

andrew8088:
This is a fix for #188.

## Cause
The bug is caused by `Eta` trying to render both static and dynamic templates when inserting a template, because it's matching the dynamic template. I think it's reading the dynamic template as `undefined + some_val`, which is why `NaN` is inserted.

## Fix
I'm fixing this by simply swapping out the opening interpolation tag with a temporary string (one unlikely to be used within the actual template), and then swapping back to the the correct tag after rendering the template.

Note: I haven't been able to run the tests (see #212), but happy to do so with a little instruction.

### DIFF
--- a/src/TemplateParser.ts
+++ b/src/TemplateParser.ts
@@ -17,6 +17,7 @@ export class TemplateParser implements TParser {
     public internalTemplateParser: InternalTemplateParser;
 	public userTemplateParser: UserTemplateParser;
     public current_context: {};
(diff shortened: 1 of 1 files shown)


SilentVoid13:
Hey @andrew8088, sorry for the late response, I've been busy for the last 2 weeks.
I fixed this in Templater 1.7.0, I ended up modifying some Eta internals to ignore dynamic commands. 
I appreciate the contribution, thanks a lot! Don't hesitate to create PRs, this helps.

Issue #204, opened 2021-05-08
Dynamic commands can't be formatted


zimt28:
**Plugin informations (please complete the following information):**
 - OS: macOS 10.15.7
 - Templater version: 1.6.0
 - Obsidian version: 0.11.13
 - Templater settings: default

**Describe the bug**
Dynamic commands don't seem to respect formatting rules:
``# <%+ tp.file.title %>``
``** <%+ tp.file.title %> **``
Both examples don't return formatted output, just the plain title.

**Expected behavior**
Formatting should be kept, the plugin probably needs to run before the markdown renderer (this would also allow something like `<%+ "# " + tp.file.title %>`).

metruzanca:
I noticed something similar with a template I'm working on that generates shields.io image links from tags, but instead of the markdown image links being rendered in the markdown they're just being treated as text.

With: 

User script: https://gist.github.com/metruzanca/71df2d19fdf29f971909a524b6853669

Obsidian Note: 
```md
---
tags:
- frontend
---
`<%+ tp.user.shields(tp) %>`
```

Expected: 
![image](https://user-images.githubusercontent.com/32832892/117697744-18ef3100-b1c3-11eb-91b2-a8e05167503b.png)

Actual: 
![image](https://user-images.githubusercontent.com/32832892/117697348-a8e0ab00-b1c2-11eb-895a-45441572d956.png)

---
@zimt28 
> the plugin probably needs to run before the markdown renderer 
yeah, I believe this is indeed the issue, as in my case its like the text is being simply appended to the markdown rendered to html.

Would be pretty awesome if Silent is able to put it before as that would open up many possibilities.

JeppeKlitgaard:
I have run into this as well. It seems to also break the metadata format:

![image](https://user-images.githubusercontent.com/1814611/119238869-97de5500-bb45-11eb-8644-2e46b049ae15.png)

SilentVoid13:
Hey @zimt28 @metruzanca @JeppeKlitgaard, thanks for the report this should now be fixed as of Templater 1.7.0.
Update and tell me if this works now for you!

JeppeKlitgaard:
@SilentVoid13 

I can't seem to get dynamic templates working in the frontmatter still. I am aware that caching is a known issue, but even using the 'tricks' that would usually update the cache doesn't make it render.

This is on 1.7.0

[...]

PR #215, opened 2021-05-12
Refactor dynamic template processing to support formatting

### LINKED ISSUE #190: Dynamic Commands inside a Table break Table formatting
 - OS: macOS
 - Templater version: 1.6.0
 - Obsidian version: 0.12.2
 - Templater settings: Template for testing  `<%+ tp.file.title %>` (chose since it's simple and output is small

**Describe the bug**
Using a dynamic command inside of a table breaks the table formatting completely and sometimes fails to render result.

**Screenshots**
image

**Additional context**
I had used another more complex template(with a few date calculations) earlier and that failed to even calculate. The template worked fine outside the table. So I tested with something simpler like just the title and it renders now but the formatting is still ruined.

andrew8088:
This resolves #190 and most of #204.

## Cause

This function originally operated on `el.innerText`, which means it would drop any child elements within `el` and only consider the text of the node. This effectively "filters out" `<h1>`, `<strong>`, `<li>`, etc.

## Fix

I've refactored this to use `el.innerHTML` instead. With `innerHTML`, the `<` and `>` characters are converted to entities when within a text node (but not when within an element attribute), so we first have to replace the entities with the correct characters. I use this opportunity to remove the `+` signs, which makes it easier to render the templates: no need to loop through the regex matching results: we can just pass the whole HTML string to the parser, and use the returned value.

There are a few outstanding issues with dynamic templates:

- Internal functions that take Obsidian links as arguments (`tp.file.include` and `tp.file.exists`) don't work, because the links are already converted to HTML when the template rules. Dynamic `include` is something I really want, and part of what prompted me to poke around, so I'll take a look at that next.
- Internal functions that output markdown will result in markdown in the final preview, instead of HTML. I think this only affects the web module functions, but this was mentioned in a comment on #204.

You can see all the rest of the internal variables and functions rendered dynamically in the screenshot below:

## Test Screenshot

Screen Shot 2021-05-11 at 10 11 06 PM

### DIFF
--- a/src/Templater.ts
+++ b/src/Templater.ts
@@ -128,35 +128,23 @@ export class Templater {
     }
 
     async process_dynamic_templates(el: HTMLElement, ctx: MarkdownPostProcessorContext): Promise<void> {
-        const dynamic_command_regex: RegExp = /(<%[*~]{0,1})\+(.*%>)/g;
+        const dynamic_open_regex: RegExp = /(<|&lt;)%([*~]{0,1})\+/g; // match both escapted and unescaped opens, so we can remove +
+        const dynamic_close_regex: RegExp = /%&gt;/g;
 
-		let content = el.innerText;
-        let match;
-        if ((match = dynamic_command_regex.exec(content)) != null) {
-			const file = this.app.metadataCache.getFirstLinkpathDest("", ctx.sourcePath);
-			if (!file || !(file instanceof TFile)) {
-				return;
-			}
-            const running_config = this.create_running_config(file, file, RunMode.DynamicProcessor);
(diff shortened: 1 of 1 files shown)


andrew8088:
Hmm, so already I've found that this doesn't play well with code blocks. It's messing with both the dataview and buttons plugins. Right now, I'm just adding this line locally

```javascript
if (el.firstElementChild.className.includes("dataview") || el.firstElementChild.className.includes("block-language-button")) return;
```

to prevent re-assigning `el.innerHTML` for those elements. But that's not a good long-term solution.

PR #275, opened 2021-07-01
Fixed modification date in the example

bartn:
Modification date in example was referencing creation_date().


SilentVoid13:
Hey @bartn, thanks for the contribution!

Issue #223, opened 2021-05-16
user function seems not to work on mobile


ebullient:
**Plugin informations (please complete the following information):**
 - OS: iOS 14.4.2
 - Templater version: 1.6.0
 - Obsidian version: 0.0.19
 - Templater settings: {"command_timeout":5,"template_folder":"[assets]/templates","templates_pairs":[["",""]],"trigger_on_file_creation":true,"enable_system_commands":false,"shell_path":"","script_folder":"[assets]/scripts"}

**Describe the bug**
This will take some explaining. Let's start with what works everywhere. I have a daily template that looks like this: 

```
<%* var titledate = tp.file.title;
const today = moment(titledate).isoWeekday();
    var nextworkday = moment(titledate).add(1, 'd');
    var thisweek = moment(titledate).day(1);
    var tomorrow = 'tomorrow';
    if (today === 1 ) {
        thisweek = moment(titledate);
    } else if (today > 4) {
        nextworkday = moment(titledate).startOf('isoWeek').add(1, 'week').day("monday");
        tomorrow = 'Monday';
    }
    tR +=  '# My Day\n'
        + moment(titledate).format("dddd, MMMM DD, YYYY") 
        + ' .... ['+tomorrow+'](' + nextworkday.format("YYYY-MM-DD")+')  \n'
        + 'Week of ['+thisweek.format("MMMM DD")+'](../weekly/Week-'+thisweek.format("YYYY-MM-DD")+')  \n';
%>
```

If I try to break this into a user function, it works on desktop, but not on mobile. Specifically, I tried replacing the template content with this:

```
<%* var titledate = tp.file.title;
const today = moment(titledate).isoWeekday();
tR += tp.user.daily(today, titledate); 
%>
```
(I use those variables elsewhere, which also seems not to work on mobile with the user function in the mix... works fine on desktop).

My user function (in the scripts folder configured above) is in a file called `daily.js`, and it looks like this:

```js
function daily(today, titledate) {
    var nextworkday = moment(titledate).add(1, 'd');
    var thisweek = moment(titledate).day(1);
    var tomorrow = 'tomorrow';
    if (today === 1 ) {
        thisweek = moment(titledate);
    } else if (today > 4) {
        nextworkday = moment(titledate).startOf('isoWeek').add(1, 'week').day("monday");
        tomorrow = 'Monday';
    }
    return '# My Day\n'
        + moment(titledate).format("dddd, MMMM DD, YYYY") 
[...]

PR #282, opened 2021-07-14
Add caution to user functions for mobile

chaseadamsio:
I ran into #223 but had to come to the GitHub issues to see that it was a problem. I added this as a temporary caution  until user functions are available.
 
image


SilentVoid13:
Hey @chaseadamsio good idea, thanks for the contribution!

PR #340, opened 2021-08-24
Handle JS user templates on mobile

shabegom:
Added handling for importing user templates when on Obsidian mobile. Instead of using the `adapter` I am reading the JS file with `app.vault.read` and then converting the returned string to a function using the `Function` constructor. because this doesn't user `module.exports` I'm stripping it from the string using regex. Changes are on lines 46-51. The other changes are formatting.


astrodad:
This would be really helpful. Any chance someone can look at this PR?

ankushg:
@SilentVoid13 Is there anything else that needs to be handled in this PR before it can be merged?

shabegom:
Closing because this has shipped via another PR

Issue #539, opened 2022-01-30
Can't import other user scripts in user scripts


Claw76:
**Plugin information:**
-   OS: Windows 10
-   Templater version: 1.10.0
-   Obsidian version: 0.13.23
-   Templater settings: Template folder location: `_templates/`, Script folder location: `_scripts/`

**Bug description:**
I import certain functions in my user scripts, which I also use as user scripts themselves. Due to the changes in Templater v1.10.0 this breaks and I get an error when importing other userscripts in a userscript. Works in Templater v1.9.11.

**Example:**
Folder structure:
_scripts:
- userscript1.js
- userscrtip2.js

userscript1.js:
```js
function f1(msg) {
 console.log(msg)
}
module.exports=f1;
```

userscript2.js:
```js
const f1 = require('./userscript1.js');
function f2(msg) {
    f1(msg);
}
module.exports = f2; 
```

**The Error:**
 > VM112:82 Templater Error: Template parsing error, aborting. 
 Cannot find module './userscript1.js'
Require stack:
\- electron/js2c/renderer_init

**What I found out so far:**
Apparently, since Templater v1.10.0, the directory where the script is "located" when executed is now:
`C:\Users\Username\AppData\Local\Obsidian\resources\electron.asar\renderer`
when up until Templater v1.9.11 it was always the `script`-folder specified in the Templater options.
Would be nice if that could be changed again.

[Test Vault.zip](https://github.com/SilentVoid13/Templater/files/7966318/Test.Vault.zip)

nicolevanderhoeven:
Thank you for this @Claw76 . I've seen this error as well, and it's breaking some of my workflows. Would love for this to be fixed.

cfurrow:
It seems that previously (in v1.9.11) the user function was parsed/loaded this way:

```javascript
const vault_path = this.app.vault.adapter.getBasePath();
const file_path = `${vault_path}/${file.path}`;
//..snip
const user_function = await import(file_path);
```

That was changed in v1.10.0:

```javascript
let req = (s: string) => {
    return window.require && window.require(s);
};
let exp: Record<string, unknown> = {};
let mod = {
    exports: exp
};

const file_content = await this.app.vault.read(file);
const wrapping_fn = window.eval("(function anonymous(require, module, exports){" + file_content + "\n})");
wrapping_fn(req, mod, exp);
const user_function = exp['default'] || mod.exports;
```

[...]

PR #543, opened 2022-02-01
Allow for `require` is User Script Functions

shabegom:
This PR fixes a regression #539 introduced in 1.10.0 that caused `require` statements to throw errors. This was because the directory in which user script functions run was no longer the script directory in the vault. We couldn't rollback this change as it would have made script functions on mobile inoperable.

**The Solution**

There is a new method called `inline_requires`. It finds `require` statements, reads contents of the required file and adds the required function inline to the user function being run. It recurses over the script until all require statements have been inlined. It also ensures duplicates are not inline by looking for the function before adding it to the final output.

This will probably add a bit of time to the overall running of user functions, but in my limited testing the impact seems negligible.


src/core/functions/user_functions/UserScriptFunctions.ts
```diff
@@ -66,7 +70,36 @@ export class UserScriptFunctions implements IGenerateObject {
...
+            let inlined = file_content;
+            if (!exists) {
+                inlined = file_content.replace(
+                    `const ${require[1]} = ${require[0]}`,
+                    require_content
+                );
+            } else {
+                inlined = file_content.replace(`const ${require[1]} = ${require[0]}`, "");
+            }
+            return this.inline_requires(inlined);
```
cfurrow, line 102: I'm running into an infinite loop here, and it seems to be happening because of lines 95-98. 

In my file, I have this:

```javascript
const extractDate = require('./date_utils.js');
```

But lines 95-98 expect it to be this:

```javascript
const date_utils = require('./date_utils.js');
```

Since my line does no look like that, the `file_content.replace` never happens, and then upon calling:

```javascript
return this.inline_requires(inlined);
```
The whole process starts over. It sees the `const extractDate = require('./date_utils.js');` line, tries to inline it, but fails to replace the `require` statement with the inline content because my line does not match up with expectations.

cfurrow (changes requested):
Thank you for doing this!

I've added some comments and suggestions that may help.

I don't have an immediate answer/suggestion for the infinite loop problem, aside from maybe needing to keep track of the full line that is going to be replaced, and not simply the name of the file.

e.g. keep track of  `const extractDate = require('./date_utils.js');` as the "key" of the inlining being done, then at the time of replacing/inlining, lookup the "key" and replace it with the inlined contents.

src/core/functions/user_functions/UserScriptFunctions.ts
```diff
@@ -66,7 +70,36 @@ export class UserScriptFunctions implements IGenerateObject {
...
+                /module\.exports = .*$/,
+                ""
+            );
+            const exists = file_content.includes(require_content);
+            let inlined = file_content;
+            if (!exists) {
+                inlined = file_content.replace(
+                    `const ${require[1]} = ${require[0]}`,
+                    require_content
+                );
```
cfurrow, line 98: I just realized you could probably do this:
```suggestion
                inlined = file_content.replace(
                    require[0],
                    require_content
                );
```

cfurrow: Well, something *like* that, not exactly that. Maybe if the require regex included the `const` as well.

It makes the regex more complicated (this is certainly not a valid regex for js variable name rules):

```javascript
/const\s+[a-zA-Z0-9\$_]+\s+=\s+require\('\.\/(.*?)(\.js)?'\)/
```

PR #576, opened 2022-03-01
fix: Activate cursor jump before renaming file

### LINKED ISSUE #564: Move cursor to a specified position after file name population (+enter)
I'm always frustrated when I enter a file name, press enter and cursor is set to header part, not to content part.
image
(at this screenshot the cursor is placed at the correct position (where content will be) but in real world it is set on the first line of the note - in this case yaml header)

After I created a new note with specific template and filled in the filename and pressed enter I want the cursor to be placed in specific template place where I specified below tag, for example:
**<% tp.file.cursor_afterfilename() %>**

**It could be one-time action, only when file is initially created (for the first time), ... when I change file name later this f-ty is not really needed...**

Right now I tried to set:
**<% tp.file.cursor() %>**
in the template, but it places cursor there immediately after note creation, not really what I wanted... I wanted this only after I enter filename first...
I also tried to find plugins which would do this but I did not succeed... 😒

**Another suitable solution would be just to place cursor at the end of the note when I press enter at filename input field. In my case it would work.** 

P.S:
this is real sample, you can see where the cursor is here (wrong)
image

liamcain:
fixes #564

This fixes a bug where the focus is stolen away from the `rename` input to place it in the document.

Instead, we set the cursor position _first_. Then we trigger the rename. After pressing <kdb>Enter</kbd> from the rename box, the cursor will return to the initial cursor_jump destination as expected.

### DIFF
--- a/src/core/Templater.ts
+++ b/src/core/Templater.ts
@@ -167,10 +167,17 @@ export class Templater {
                 return;
             }
             await active_leaf.openFile(created_note, {
-                state: { mode: "source" },
-                eState: { rename: "all" },
+              state: { mode: "source" },
             });
-            await this.plugin.editor_handler.jump_to_next_cursor_location(created_note, true);
+
+            await this.plugin.editor_handler.jump_to_next_cursor_location(
+              created_note,
+              true
+            );
+
+            active_leaf.setEphemeralState({
+              rename: "all",
+            });
         }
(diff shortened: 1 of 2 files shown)


tangor86:
It seems not fixed... please see gif
cursorWrongPlacement

Issue #550, opened 2022-02-06
User Scripts silently fail on mobile in 1.10.0


kevboh:
**Plugin informations (please complete the following information):**

-   OS: iOS 15
-   Templater version: 1.10.0
-   Obsidian version: 1.0.5
-   Templater settings: `templates/` folder, `templates/js` user scripts folder

**Describe the bug**

When creating or inserting a template on mobile, if that template contains a user script the template will be inserted without the template logic running. No logs appear (per mobile console in dev tools plugin).

Steps to reproduce:

Add the following user script as nothing.js in your scripts folder:

```js
function nothing() {
  // do nothing
  return "hello world";
}
module.exports = nothing;
```

It works on desktop, and as a CustomJS script. Based on my experience this function can be anything, but this is one I can reproduce with.

Next, add a template that calls the user script:

```md
<% tp.user.nothing() %>

Some other text
```

Insert the template into an empty note on desktop. It will work. Insert on mobile. It will not.

**Expected behavior**
Should work the same on desktop and mobile.

kevboh:
In case it's useful: I was able to work around this issue with my scripts by moving them to [CustomJS](https://github.com/SamLewis0602/obsidian-custom-js), then calling them per the example in CustomJS' readme. So I don't think this is an Eta issue; it seems to be something with script calling specifically. The user scripts are listed on mobile in the settings panel.

sekwah41:
I'm not sure why the scripts were disabled in the first place, if you force the scripts to load it all seems to run fine for all the cases I am using at least.

Id rather it not work with an error message in the cases people are breaking it than disable it entirely for everyone if its still an issue.

PR #619, opened 2022-04-17
Allow User Scripts on Mobile

### LINKED ISSUE #586: User Script Functions are not working on mobile
**Plugin informations:**

-   OS: Android 
-   Templater version: 1.11.3
-   Obsidian version: 1.1.1 (44)
-   Templater settings: 
    - Template: "10 Meta/Templates/testing.md" 
    - User function: "10 Meta/Templates/Scripts/notice.js"

Content of the **template** (testing.md):
```
<%* tp.user.notice(tp) %>
```

Content of the **script** (notice.js):
```js
async function notice(tp) {
    const text = await tp.system.prompt("What's Good?")
    new Notice(text, 5000)
}

module.exports = notice
```

**Describe the bug**
The function doesn't run with the following error:
```
[error] Templater Error: Template parsing error, aborting. 
 tp.user.notice is not a function
```

**Expected behavior**
The user function runs, displaying a prompt.

**Additional context**
- I have tested this with other scripts as well, none work.
- The template folder is set to `10 Meta/Templates`
- The script folder is set to `10 Meta/Templates/Scripts`
- The script is recognised by Tenplater in the settings:

***
Thanks for the great work with Templater. I'd love to use user functions on mobile too, but always run into an error.

pjkaufman:
There was an issue where if you tried to use user scripts on mobile devices, it would throw an error saying that your script did not exist. It would still detect your user scripts in the settings, but you would not be able to generate any templates. Removing the check to only allow user scripts for desktop seems to have resolved the issue.

This is working for me on my personal Android phone. I am not sure which mobile device caused this to be solely allowed for desktop prior, but it may be that Obsidian has been updated in such a way that the issue has gone away.

fix #586

Changes Made:
- Removed Platform import from Obsidian and using it to check if the current platform is desktop in `UserFunctions.ts`

### DIFF
--- a/src/core/functions/user_functions/UserFunctions.ts
+++ b/src/core/functions/user_functions/UserFunctions.ts
@@ -1,4 +1,4 @@
-import { App, Platform } from "obsidian";
+import { App } from "obsidian";
 
 import TemplaterPlugin from "main";
 import { RunningConfig } from "core/Templater";
@@ -26,9 +26,8 @@ export class UserFunctions implements IGenerateObject {
                 await this.user_system_functions.generate_object(config);
         }
 
-        // TODO: Add mobile support
         // user_scripts_folder needs to be explicitly set to '/' to query from root
-        if (Platform.isDesktopApp && this.plugin.settings.user_scripts_folder) {
+        if (this.plugin.settings.user_scripts_folder) {
             user_script_functions =
                 await this.user_script_functions.generate_object(config);
         }


pjkaufman:
@SilentVoid13 , I will await your review. I am not sure how we are supposed to run the UTs since I was unable to do so by running any of the `npm run ...` commands. Please let me know if you would like changes or anything like that.

sekwah41:
Can confirm this edit works on an iPad.

I can also check on other platforms if needed.

Could we possibly get a link to the initial issue and why this was disabled? 

It seems that it may have not worked in a few cases with imports or something but for things like utility functions which just wrap checking dates and most of what I use it for it works perfect.

Id rather this be enabled with the error messages than entirely remove it for those who don't need whatever cases it was erroring in. Unless it was possibly causing larger data loss somehow.

sekwah41:
I believe this also relates to issue #550 too

shabegom:
merging

martijnarts:
Thanks @shabegom!

Issue #647, opened 2022-05-27
Add templater ribbon/sidebar button toggle


fmbarina:
**Is your feature request related to a problem? Please describe.**
I never use the ribbon/sidebar icon, so I wish I could just remove it to make things look tidier.

**Describe the solution you'd like**
A toggle inside the plugin options to hide/show the ribbon icon.
I think every plugin with a ribbon icon would do well to have this, to be honest.

**Describe alternatives you've considered**
- Pretty sure it's possible with custom CSS.
- phibr0/obsidian-customizable-sidebar has a "Hide Templater?" option.
- Disabling Templater removes the icon.

But these solutions bring small downsides, such as needing external plugins or being somwhat inconvenient.

If you have any ideas or opinions regarding this, I'd love to hear them.

I tried writing the code myself, and it seems to work. I do not have experience with this kind of thing. 
May I open a PR for the implementation of this feature?

welpdx:
That is a good request! I have thought about this but I totally forgot about it. 
After a bit of fiddling, this is what worked for me:

1) You can use this to hide all of those action ribbons:
```js
<%* this.app.workspace.leftRibbon.ribbonActionsEl.hidden = true %>
```
2) Or you can use this to hide the templater ribbon only:
```js
<%* 
var actionRibbons = this.app.workspace.leftRibbon.ribbonActionsEl.childNodes
for (var i = 0; i < actionRibbons.length; i++) {
    if (actionRibbons[i].ariaLabel == "Templater") {
        actionRibbons[i].hidden = true
    }
} 
%>
```
You can add to it to hide other ribbons as well. You can add this to a Templater Startup Template so these ribbons can be hidden on startup. 

Hope that helps. 

~Welp

fmbarina:
Thank you, I hadn't even thought about using a startup template. 
I like that this way, the problem is solved with Templater itself and no changes are required.

Despite that, I still think a toggle in the plugin options would be better though, since:
- There's no need to use a startup template, it's just one click in the options.
- Other people who might want this will benefit without having to find this issue.

I hope I'm not being too demanding about this, this is just my honest opinion. I've seen some PRs where people just offer the enhancement, no issue needed, so I was wondering if I should try making one as well.

welpdx:
Hey @fmbarina, it is not demanding at all. In fact, I think that user feedback and user requests is what makes obsidian, templater, and softwares in general, better. 

[...]

PR #653, opened 2022-05-30
Add option to show/hide Templater ribbon icon

fmbarina:
Enhancement #647 I suggested recently.

Add a plugin option to show or hide the sidebar/ribbon icon. There are other ways, including even [another obsidian extension](https://github.com/phibr0/obsidian-customizable-sidebar) with an option to hide the Templater icon. I think it makes more sense and is better for the extension itself to offer that option.

Changes Made:
- Added boolean setting enable_ribbon_icon, default true, toggle option hides/shows icon.
- Added check to see if setting is true before adding icon in main.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -39,9 +39,11 @@ export default class TemplaterPlugin extends Plugin {
         this.command_handler.setup();
 
         addIcon("templater-icon", ICON_DATA);
-        this.addRibbonIcon("templater-icon", "Templater", async () => {
-            this.fuzzy_suggester.insert_template();
-        });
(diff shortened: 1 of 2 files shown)


shabegom (changes requested):
just needs one small language tweak

fmbarina:
Thanks, changed. Really makes me wonder if I can truly call myself fluent.
Squashed into a single commit so that it fits better in the commit history. I think that's how you're supposed to do it?

shabegom:
Sweet, thanks! Will merge and add to next release

PR #667, opened 2022-06-18
Updating examples to clarify and simplify

patrickkelso:
I spent a lot of time trying to figure out the syntax for using throw_on_cancel because there was no example added when the feature was. 

I've also updated the examples so each one explains a little of what it does making it easier to identify a use case. I think this will make the documentation a lot more accessible to non programmers.

### DIFF
--- a/docs/src/internal-functions/internal-modules/system-module.md
+++ b/docs/src/internal-functions/internal-modules/system-module.md
@@ -31,13 +31,27 @@ Function documentation is using a specific syntax. More information [here](../..
 {%- endfor %}
 
 ## Examples
-
+### Paste from clipboard
 ```javascript
 Clipboard content: <% tp.system.clipboard() %>
-
+```
+### Standard prompt, will return null if cancelled.
+```javascript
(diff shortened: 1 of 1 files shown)


AB1908:
Hey while you're on this, can you modify the docs to make the examples be part of the function's section instead of being jumbled together at the end?

shabegom:
The docs should be generated on release. @patrickkelso please add these examples to the `documentation.toml` file under the appropriate header. example:

```
[tp.system.functions.prompt]
name = "prompt"
description = "Spawns a prompt modal and returns the user's input."
definition = "tp.system.prompt(prompt_text?: string, default_value?: string, throw_on_cancel: boolean = false)"
example = "<% tp.system.prompt("Please enter a value") %>"
```

shabegom (changes requested):
Move examples under appropriate functions in documentation.toml

patrickkelso:
Understood. I'll update asap.

Zachatoo:
Doing some cleanup, please reopen if you'd still like to contribute!

PR #697, opened 2022-07-18
Feature: multiline input with `tp.system.prompt`

AB1908:
I've added a new multi-line input prompt inspired by chhoumann/QuickAdd. See [here](https://github.com/chhoumann/quickadd/blob/master/src/gui/GenericWideInputPrompt/GenericWideInputPrompt.ts) for original.

It can be invoked using
```js
let default_value = "Original";
let prompt_text = "Prompt Title Here";
let throw_on_cancel = false;
let multi_line = true;
tR += await tp.system.prompt(prompt_text, default_value, throw_on_cancel, multi_line);
```

Sample:
image

This is an optional variable that shouldn't introduce any breaking changes. I haven't had a chance to test this on mobile.

Note: the event handler for mobile behaves slightly differently. For multi-line inputs, the Submit button must be used as the enter key moves the cursor to the next line.

Please let me know if there's any cleanup I should do.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1,7 +1,7 @@
 {
     "id": "templater-obsidian",
     "name": "Templater",
-    "version": "1.12.0",
+    "version": "1.13.0",
     "description": "Create and use templates",
     "minAppVersion": "0.11.13",
     "author": "SilentVoid",
--- a/manifest.json
+++ b/manifest.json
@@ -1,7 +1,7 @@
 {
     "id": "templater-obsidian",
     "name": "Templater",
-    "version": "1.13.0",
+    "version": "1.12.0",
     "description": "Create and use templates",
     "minAppVersion": "0.11.13",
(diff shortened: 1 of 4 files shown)


AB1908:
If anyone needs this, feel free to install 1.13.0 from my fork via BRAT.

manifest.json
```diff
@@ -1,7 +1,7 @@
 {
     "id": "templater-obsidian",
     "name": "Templater",
-    "version": "1.12.0",
+    "version": "1.13.0",
```
shabegom, line 4: don't update the version number in a feature PR, we'll do that when cutting a new release candidate. remove this from the PR.

shabegom (changes requested):
remove manifest.json

AB1908:
I did that for the brat thing. I'll split that off into a separate branch next time.

shabegom (approved):

Issue #684, opened 2022-07-05
Support front matter creation when using DB Folder


Mattrogers13:
Currently when using Templater with the DB Folder plugin the DB Folder creates Front Matter first within a new note causing Templater not to trigger.

I would like to have an option to still trigger when file size is greater than 0 or when front matter is detected. 

Thank you!

AB1908:
Oh if this is what you want, I think you can throw this in a PR. It's unlikely anyone will get to this soon.

Mattrogers13:
ok, will do. Thanks.

Mattrogers13:
I now realize I do not understand how to add this to a PR.

welpdx:
That's ok. A PR is a Pull Request. 
If know how to fix this issue, you can fork this repo, add you changes, and then do a pull request. The PR is a request for @SilentVoid13 to "pull" your changes from your fork to his repo.

TheAussieTinker:
I don't know how to do a PR but if you remove `file.stat.size == 0 ` on line 3886 then it works with DB Folder

TheAussieTinker:
If this breaks other workflows would it be possible to add a setting that turns off this check for the Data Base Folder users.

Sklore:
Oh yes please this functionality would be awesome! People are trying to achieve [THIS](https://youtu.be/ZKOYtrVXSbk). 
I would be a bit nervous about just taking this line out with insight from @SilentVoid13 though. I assume it's there for a reason and would not want to break anyone else's workflows. 
Ideally, possibly a setting so you could turn this on/off might be a better solution.

PR #713, opened 2022-08-03
#684 - Remove 'file.stat.size == 0' from line 394

ObsidianTTRPGProject:
This line of code means that new files created with the DataBase Folder Plugin do not trigger Templater. Lots of people are trying to use DataBase Folder and Templater together so this pull removes the file size check which enables Templater to trigger.

Related to issue #684

### DIFF
--- a/src/core/Templater.ts
+++ b/src/core/Templater.ts
@@ -391,7 +391,6 @@ export class Templater {
         await delay(300);
 
         if (
-            file.stat.size == 0 &&
             templater.plugin.settings.enable_folder_templates
         ) {
             const folder_template_match =


liamcain:
Removing this line will mean that Templater will trigger when files get moved into a new folder. This will have some dangerous consequences.

AB1908:
Can we listen precisely for new file events instead or do file moves also trigger that?

liamcain:
A file move will always trigger that; even if it's something we tweaked at the API-level, you could just as easily move a file from your operating system's file system and obsidian won't be able to tell if it's a newly created file or a file that was moved into the vault.

AB1908:
I know I'm asking a lot here but do you see any reasonable trade-off here? I don't really have the depth to figure out the right approach here. If you can't, I'll try and block some time to investigate this.

liamcain:
So, a few options off the top of my head. 

1. The DataBase Folder Plugin could just check if templater is installed, and if it is, directly trigger a command to apply templates based on the file location. This is the easiest IMO but maybe not the most kosher.

2. The Templater plugin could introduce a new event on the app workspace (for example: `new-markdown-file-created` and the DataBase Folder Plugin could call `workspace.trigger('new-markdown-file-created')` after it creates a new file. This would be roughly equivalent to option 1. but slightly more robust since it doesn't require checking if another plugin is installed.

3. (worst option, don't do this.) DataBase Folder Plugin can create an empty file, wait some amount of time for templater to modify it, then it can add the frontmatter afterwards.

---

Some other random ideas that I don't think are very viable but jotting down for completeness.

- Some frontmatter flag that db folder plugin can include to mark the file as ready to be processed to templater. This could easily lead to misuse or bugs so don't do this either.
- Checking the ctime to see if the file is new vs moved. This will lead to issues if the file was moved via a cloud syncing solution or if the ctime ever got reset programmatically

There are definitely better approaches and I'll give it some thought for how this can be solved from Obsidian's side. Templater's Folder Templates are pretty dangerous IMO and I think we can offer something better if it's baked into the core app.

Issue #691, opened 2022-07-16
New note replaces File Explorer panel when using Create new note from template


codyburleson:
This never happened before, but now, in the new version of Obsidian, when I choose to create a new note from template, the new note is created right there in place of the File Explorer tab. I have to close and reopen Obsidian in order to get the file explorer tab back. This is Obsidian version 0.15.6 running on Windows 10.

akay64:
Super annoying bug, I think I will try to roll back to the previous version of Obsidian till this gets worked out.

AB1908:
Hey Cody! Nice to see ya after a while. I think you're capable of fixing this actually. `app.workspace.activeLeaf` behavior changed with 15.4 and we need to handle it a little differently. You can find an example in my PR for obsidian-spaced-repetition. In case you can't get to it, I'll try and push a fix in a few days.

welpdx:
Similar to (#670). I currently use this workaround by adding it to my templates. 

```js
<%*
if (this.app.workspace.activeLeaf.getViewState()?.type =="file-explorer") {
this.app.workspace.setActiveLeaf(this.app.workspace.getMostRecentLeaf())
}
%>
```

Or open command palette and run `Reload app without saving` to get your File Manager back.

welpdx:
@AB1908  is this the PR you were talking about? https://github.com/st3v3nmw/obsidian-spaced-repetition/pull/469
Pretty comprehensive

AB1908:
Yep, that's the one.

AB1908:
I guess I'll get around to working on this but no guarantees on the timeframe.

AB1908:
Haha so many contributions, hard to stay on top of it all.

yummy527:
I also have this issue on arch linux. Obsidian version: v0.15.6.

DR3AIVI:
I have this problem as well on 0.15.9  Another work around is to right click on the tab of the new note and click "Reveal file in navigation"

codyburleson:
I am able to fox this by changing the `const active_leaf =` line in the `if(open_new_note)` block in `Templater.ts`. See BEFORE commented out and AFTER as the line just after...

```js
if (open_new_note) {
            // const active_leaf = this.app.workspace.activeLeaf;
            const active_leaf = yield this.app.workspace.getLeaf(false);
            if (!active_leaf) {
                log_error(new TemplaterError("No active leaf"));
                return;
            }
```

[...]

PR #736, opened 2022-08-27
fix issue #691

codyburleson:
Instead of getting the activeLeaf from the workspace, which is the File Explorer when you're right-clicking to create a new file from a template, this gets the active leaf from the root split, which is more commonly where you want to be editing. This fix has been validated by myself and two others.


AB1908 (approved):

shabegom:
merged without the `yield`

PR #745, opened 2022-09-01
delete question template from Issues & remove auto-assign to SilentVoid

shabegom:
I think questions need to be asked in the Discussions feature and issues should be limited to feature requests and bug reports only.

Also I don't think we should auto-assign all issues to SilentVoid


shabegom:
I'm merging this.

Issue #595, opened 2022-03-20
Template fails to load if it contains "web" commands when internet is unavailable


Lurux:
- OS: Linux Mint 20.3 ≈ Ubuntu 20.04
- Templater version: 1.12.0
- Obsidian version: 0.13.33
- Templater settings: using `tp.web.random_picture()` or `tp.web.daily_quote()` as part of a template

<details><summary>Template code</summary>

```
[[Todo]] - [[Daily/<% tp.date.now("D-MM-Y") %> | Daily note]]


**Random notes**

<%*
let notes = app.vault.getMarkdownFiles();

for(let i = 0; i < 5; i++) {
	const index = Math.floor(Math.random() * notes.length);
	tR += `\n- [[${ notes[index].basename }]]`;
}
%>

**Random quote**

<% tp.web.daily_quote() %>

**Random image**

<% tp.web.random_picture() %>
```
</details>

**Describe the bug**

When trying to load that template while internet is disconnected or unavailable, the whole template fails to load with message "template parsing error, aborting"

<details><summary>Console errors</summary>

```
GET https://api.quotable.io/random net::ERR_INTERNET_DISCONNECTED
Templater Error: Template parsing error, aborting. Failed to fetch
GET https://source.unsplash.com/random/? net::ERR_INTERNET_DISCONNECTED
```
</details>

(Do note the template works as expected without internet if I remove the two web commands)

**Expected behavior**

The template should be able to load normally, with only the two "web" commands either returning blank text or printing an error to the file.

**Additional context**

(I think everything's up there)

welpdx:
You can put those function in a try catch, to ensure that your whole script runs all the way to the end. 
```
<%*
    try {
        tR+= await tp.web.random_picture()
    } catch (err) {
        tR+= err // display error
    }
%>
```


When there is no internet, the error is `TypeError: Failed to fetch`. You can comment out `tR+= err` in the the catch block if you do not want to display this error. 

Hope that helps.

~Welp

AB1908:
I'd recommend creating a `new Notice()` instead of adding to `tR`.

welpdx:
> I'd recommend creating a `new Notice()` instead of adding to `tR`.

The above code meets what the guy wanted 
> either returning blank text or printing an error to the file.


But I like your idea. 

```
<%*
    try {
        tR+= await tp.web.random_picture()
    } catch (err) {
        new Notice("Error: \n" + err , 2000);
    }
%>
```

PR #785, opened 2022-09-02
fix: add try/catch logic to web commands

shabegom:
Add try/catch to:
- daily quote
- random image

This should prevent them from causing full pages to fail.


shabegom:
Tested and seems to be working

welpdx:
Fixes #595 #779

PR #793, opened 2022-09-04
Revert "feat: Update sizing of image"

wernersa:
After reviewing unintended effects. See https://github.com/SilentVoid13/Templater/issues/792

This reverts commit dbc3df55647e4dd2167e0ae47dc161b4e181dc0f.

Issue #792, opened 2022-09-03
Breaking Change: Random Picture Size


ManadayM:
**Plugin information (please complete the following information):**

-   OS: Mac OS 10.15.7
-   Templater version: 1.13.0
-   Obsidian version: 0.15.9
-   Templater settings: 

**Describe the bug**
The latest Templater release breaks the ideal behavior of the max image width when a user has enabled the "Readable line length" Obsidian setting (Obsidian Settings > Editor > Display > Readable line length). The generated random image now occupies the actual width which has been set in the `random_picture(900x300)` function. 

I downgraded the plugin with 1.12.0 by manually downloading the assets and it restored the expected behavior.

**Expected behavior**
The function should respect the readable line length setting i.e. max container width set by the Obsidian.

**Screenshots**
My default note template contains `<% tp.web.random_picture("900x300", "tree,landscape,water,mountain") %>` on top.

In release 1.12.0 - Respects the Obsidian setting
![Screenshot 2022-09-03 at 17 45 41](https://user-images.githubusercontent.com/521134/188270007-5d6b9df5-3b3d-48e0-a4e7-0b5e483081b3.png)

In release 1.13.0 - Ignores the Obsidian setting
![Screenshot 2022-09-03 at 17 48 48](https://user-images.githubusercontent.com/521134/188270147-1f47f073-54b4-4561-982b-47545a70bc17.png)

**Additional context**
The following code change in the latest release is breaking this behavior.

https://github.com/SilentVoid13/Templater/pull/678/files#diff-74b11ab27692d63b0d5b5bffffb9ad72bcd6a5a5c8b6ea21dbca9fa61aad9495L52

Hope it helps. Happy to help.

Thanks,
Manaday

shabegom:
Great report. Will look into it

shabegom:
@wernersa any thoughts here? My first instinct is to revert back to the previous way as it didn't seem to impact image display.

welpdx:
Can't replicate with:
Windows OS
Obsidian Version: 0.16.1
Templater Version: 1.12.0

wernersa:
Problem goes both ways here as not having the option for manually setting the display values of images can interfere with several plugins and themes that modify image display; specifically the most used theme "Minimal". This is why I suggested the initial change. See the following sandbox where I set theme to Minimal:

[...]

PR #794, opened 2022-09-04
Fix for web rand img resizing bug. See #792

welpdx:
Fix unintended effects (i.e., not respecting previewer and editor padding). See #792
This uses html tag instead of markdown from dbc3df5.
Alternative fix to #793.


shabegom:
I prefer this fix to a revert. Only worry is that people will not be as comfortable with the html syntax versus markdown.

welpdx:
I agree. It functions as intended now but it is no longer markdown.

shabegom:
closing in favor or #810. Thanks!

PR #802, opened 2022-09-05
fix: prompt default_value and placeholder order

### LINKED ISSUE #801: default_value doesn't set value in `tp.system.prompt` after update

-   OS: [e.g. Windows 10]
-   Templater version: 1.13.0
-   Obsidian version: 0.15.9

**Describe the bug**
After recent update (I'm not sure if it's Obsidian update or plugin update), the `tp.system.prompt` no longer has correct behavior for `default_value`. `default_value` correctly shows in the input prompt, but does not get set to the variable unless user makes some change to the value in the input field. Particularly, when `default_value` is unchanged and user simply clicks "submit" button, the value is left blank, rather than being the value originally set in `default_value`. In other words, it is almost as if it is a placeholder rather than a default value. 

**Expected behavior**
`default_value`, when unchanged, should be correctly returned from the function. 

Thanks so much, great plugin!

shabegom:
For some reason, the prompt has a placeholder and a value set. In the last release, the placeholder was set after the value, overwriting any default_value set by the user. This PR changes the order back.

fixes #801

### DIFF
--- a/src/core/functions/internal_functions/system/PromptModal.ts
+++ b/src/core/functions/internal_functions/system/PromptModal.ts
@@ -1,5 +1,12 @@
 import { TemplaterError } from "utils/Error";
-import { App, ButtonComponent, Modal, Platform, TextAreaComponent, TextComponent } from "obsidian";
+import {
+    App,
+    ButtonComponent,
+    Modal,
+    Platform,
+    TextAreaComponent,
+    TextComponent,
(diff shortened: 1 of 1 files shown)


AB1908:
Oh that's what it was. Sorry I didn't get to it sooner, and thanks for cleaning up after me 😅

PR #813, opened 2022-09-07
Fix PromptModal default value needs to be set to this.value on load

GitMurf:
Issue where it is undefined if you just accept the default string. Happens with all combo of options multi line true/false... error throw true/false. See video.

https://user-images.githubusercontent.com/64155612/189000550-c201fb6c-01ed-4d26-8a9f-eab062308a72.mp4

### DIFF
--- a/src/core/functions/internal_functions/system/PromptModal.ts
+++ b/src/core/functions/internal_functions/system/PromptModal.ts
@@ -56,9 +56,10 @@ export class PromptModal extends Modal {
             textInput = new TextComponent(div);
         }
 
+        this.value = this.default_value ?? "";
         textInput.inputEl.addClass("templater-prompt-input");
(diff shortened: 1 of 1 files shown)


GitMurf:
@shabegom I will refactor the code a bit as discussed in Discord to simplify it instead of using the `onchange` but wanted to fix the bug first. Let me know if you have any questions. I tested against all my scenarios that were breaking and they all work now.

shabegom (approved):
LGTM

shabegom:
@GitMurf for future PRs, please follow conventional commits spec:https://conventionalcommits.org/

GitMurf:
@shabegom just so i’m clear, my commit message was fine except needed to add a “:” colon and make Fix lowercase? Anything else?

Issue #812, opened 2022-09-07
PR #444 changed behavior of file.move


ebullient:
1) Mobile no longer works due to reliance on app vault (it worked fine on mobile before this change): 
```
// TODO: Add mobile support
if (Platform.isMobileApp) {
    return UNSUPPORTED_MOBILE_TEMPLATE;
}
if (!(this.app.vault.adapter instanceof FileSystemAdapter)) {
    throw new TemplaterError(
        "app.vault is not a FileSystemAdapter instance"
    );
}
```

2) The mkdir capability is not correctly handling nested directories. Given path A/B/C/filename, I am seeing a phantom directory C created at the root level due to this change:

```
this.app.vault.adapter.mkdir(dirname(new_path).split(sep).pop());
```

shabegom:
Maybe because of the `path` import... can we use `path` on mobile?

shabegom:
@alchemistake, feel like taking a look at this?

PR #444, opened 2021-11-15
Additional argument for tp.file.move for passing down specific files.

alchemistake:
# What
This adds an additional optional parameter for moving any `TFile` inside the vault. With this edit, existing user templates does not break since parameter is optional and It gives more flexibility if some people want to pass down a file explicitly.

# Why
I had a problem with `tp.file.move()` because of `this.config.target_file` being different in the different scopes of my template. 

For example:
```
<%
tp.file.create_new("", "a_file_I_dont_want_to_move);
console.log(this.config.target_file);
tp.file.move("new_location/my_file");
%>
```
This code snippet moves the newly generated file rather than the file I triggered the template with. Console log outputs the file I want to move a.k.a. active_file.

Also updated the doc to have the optional parameter, if it is not clear I can add more details there.

closes #341

PR #814, opened 2022-09-08
fix: move on mobile using obsidian api (not path)

ebullient:
Repairs regression to move support on mobile added by #444.

Preserves behavior of second parameter, and ensures target directory exists first.


shabegom (approved):
LGTM

shabegom:
Fixes #812

PR #816, opened 2022-09-10
fix: don't trigger on creation unless new file is opened in a pane

### LINKED ISSUE #554: Pause sync while replacing templates in a note
Having obsidian opened on 2 devices, when I create a note with a template prompting for inputs or suggester, the prompts are shown on the two devices.
This is because I have "Trigger Templater on new file creation" selected: therefore my new note being created with a template on device A is also synced on device B before I can complete all the inputs on device A;
each device triggers templater at the same time and the second inputs that I complete on device B overrides the inputs that I've completed on device A.

Put an option in setting tab to "pause" the sync while replacing the template in a note and put it on "resume" when templater has completed the template replacement.
Manage the case when an error occurs while replacing the template to avoid the sync staying in "pause"

Close Obsidian on device B when triggering Templater ;) sure not the best UX

**Additional context**
Thank you so much for your plugin, this is core in my automation process

### LINKED ISSUE #716: 3rd party syncing triggers new file fill-in in templater folder
**Plugin information:**

-   OS: Windows 10
-   Templater version: 1.12.0
-   Obsidian version: v0.15.9
-   Templater settings:
    - Template folder location: "9 extras/90090 hidden!/hidden! 8 templates/hidden! 801 templater", 
    - Trigger Templater on new file creation: true

**Describe the bug**
google drive being used for syncing vault across multiple devices triggers new file creation when syncing new template files.

**Expected behavior**
[during syncing, or anytime] automatically added template files in the designated templater folder are ignored by new file trigger.

**Additional context**
unintentional population of all of my template files has happened 3 times to me now, usually because a folder is renamed and then synced to my other computer, where the files are promptly populated and then synced back to the cloud all in a few seconds. i had to go through each file in file recovery and paste their old contents all in 30 minutes.
i am not sure if it is possible to differentiate whether a new file is from a syncing or not, which is why i suggested the expected behavior to be ignoring newly added files in the templater folder, but really any safety measure feature for this would be helpful.

b0o:
See my comment from https://github.com/SilentVoid13/Templater/issues/716#issuecomment-1242573930

- I've done a bit of testing but would appreciate others verify that this doesn't break anything. I am not familiar with this codebase.
- I chose `300ms` as the max threshold between file creation and opening - works on my machine, but not sure if it might break on slower machines?
- I wonder if a similar event+timeout approach could be used to eliminate the need for the `await delay(300)`

Fixes #716, fixes #554

### DIFF
--- a/src/core/Templater.ts
+++ b/src/core/Templater.ts
@@ -399,35 +399,53 @@ export class Templater {
             return;
         }
 
-        // TODO: find a better way to do this
-        // Currently, I have to wait for the daily note plugin to add the file content before replacing
-        // Not a problem with Calendar however since it creates the file with the existing content
-        await delay(300);
+        await new Promise<void>((resolve) => {
+            let timeout: NodeJS.Timeout;
+            const onFileOpenRef = templater.app.workspace.on("file-open", async (openedFile: TAbstractFile) => {
+                if (openedFile !== file) {
+                    return;
+                }
+                clearTimeout(timeout);
+                templater.app.workspace.offref(onFileOpenRef);
 
-        if (
-            file.stat.size == 0 &&
-            templater.plugin.settings.enable_folder_templates
-        ) {
-            const folder_template_match =
-                templater.get_new_file_template_for_folder(file.parent);
-            if (!folder_template_match) {
-                return;
-            }
+                // TODO: find a better way to do this
+                // Currently, I have to wait for the daily note plugin to add the file content before replacing
+                // Not a problem with Calendar however since it creates the file with the existing content
+                await delay(300);
 
-            const template_file: TFile = await errorWrapper(
-                async (): Promise<TFile> => {
-                    return resolve_tfile(templater.app, folder_template_match);
-                },
-                `Couldn't find template ${folder_template_match}`
-            );
-            // errorWrapper failed
-            if (template_file == null) {
-                return;
-            }
-            await templater.write_template_to_file(template_file, file);
-        } else {
-            await templater.overwrite_file_commands(file);
-        }
+                if (
+                    file.stat.size == 0 &&
+                    templater.plugin.settings.enable_folder_templates
+                ) {
+                    const folder_template_match =
+                        templater.get_new_file_template_for_folder(file.parent);
+                    if (!folder_template_match) {
+                        resolve();
+                        return;
+                    }
+
+                    const template_file: TFile = await errorWrapper(
+                        async (): Promise<TFile> => {
+                            return resolve_tfile(templater.app, folder_template_match);
+                        },
+                        `Couldn't find template ${folder_template_match}`
+                    );
+                    // errorWrapper failed
+                    if (template_file == null) {
+                        resolve();
+                        return;
+                    }
+                    await templater.write_template_to_file(template_file, file);
+                } else {
+                    await templater.overwrite_file_commands(file);
+                }
+                resolve();
+            });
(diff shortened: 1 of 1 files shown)


shabegom:
Thanks for the PR. Will review tonight or tomorrow.

src/core/Templater.ts
```diff
@@ -399,35 +399,53 @@ export class Templater {
...
-            file.stat.size == 0 &&
-            templater.plugin.settings.enable_folder_templates
-        ) {
-            const folder_template_match =
-                templater.get_new_file_template_for_folder(file.parent);
-            if (!folder_template_match) {
-                return;
-            }
+        await new Promise<void>((resolve) => {
+            let timeout: NodeJS.Timeout;
```
shabegom, line 403: `NodeJS.Timeout` should just be `Timeout` since we are running in a browser context and not a node context

src/core/Templater.ts
```diff
@@ -399,35 +399,53 @@ export class Templater {
...
+            const onFileOpenRef = templater.app.workspace.on("file-open", async (openedFile: TAbstractFile) => {
```
shabegom, line 404: `openedFile` should be type `TFile`

src/core/Templater.ts
```diff
@@ -399,35 +399,53 @@ export class Templater {
...
+                        resolve();
+                        return;
+                    }
+                    await templater.write_template_to_file(template_file, file);
+                } else {
+                    await templater.overwrite_file_commands(file);
+                }
+                resolve();
+            });
+            timeout = setTimeout(() => {
```
shabegom, line 444: I wonder if we update the `delay` utility function to take an optional function as a second argument. then this could be `delay(300, timeOutFunc)`. Not a need to have, but might be a useful abstraction for other parts of the codebase... could come in a separate PR.

shabegom (changes requested):
Looks good! Functional testing showed it worked as expected and there are only a couple of typing changes requested. The typings in this plugin are all kinds of messed up, so I'm going to try and fix those next.

I'm going to open a discussion in the Templater discover channel on OMG to make sure we're not missing any edge cases where you want the template to fire without the file being open.

src/core/Templater.ts
b0o: Nice catch, though [it seems it should just be `number`](https://github.com/microsoft/TypeScript/blob/12ab0fea9f45c48ea644c8273bedde235c8bcc61/lib/lib.dom.d.ts#L18048) in the browser context.

b0o: P.s. do you know if there is a good reason `@types/node` is included in devDeps? It's confusing my editor unless I explicitly prefix as `window.setTimeout` / `window.clearTimeout`.

b0o: The Obsidian API definition file actually [recommends](https://github.com/obsidianmd/obsidian-api/blob/6b2138aa28fb6d6f58f78e6fabec02186fdc0f82/obsidian.d.ts#L666) using `window.` in a similar case to avoid TS getting confused:

```ts
Use {@link window.setInterval} instead of {@link setInterval} to avoid TypeScript confusing between NodeJS vs Browser API
```

shabegom: The typings in this plugin are really messed up. I opened a PR today to fix them. In other plugins I've used `Timeout` without issue. `window.` is fine too.

shabegom (approved):
Will merge and release this weekend

Issue #728, opened 2022-08-22
templater script fails if comment contains apostrophe


Smitty010:
I'm using the latest obsidian and templater plugin (v0.15.9/1.12.0).  I saw this bug which seemed really odd to me. inside the templater script, I had a comment which contained the word "doesn't".  

Here's how I just replicated the problem.  I added this to a template.

```
<%*
// doesn't work
%>
```

I get the log message
```
plugin:templater-obsidian:82 Templater Error: Template parsing error, aborting. 
 unclosed string at line 62 col 9:

  // doesn't work
          ^
```
Removing the comment makes the problem go away.  Is this expected behavior?  Is there some preprocessing on the script that doesn't ignore comments when doing an initial scan?

shabegom:
What happens if you escape the apostrophe: doesn\'t

I don't think there is any special handling of comments when processing a command.

SilentVoid13:
Hey @Smitty010 I think this was fixed in 1.6.0, could you update and confirm?

Smitty010:
I assume you mean templater 1.16.0.

I think it may be fixed.  I modified a templater file I had to add a comment that had an apostrophe.  I got a syntax error.  I then updated.  I still got a syntax error.  However, I realized that I screwed up adding the comment in in such a way that there really was a syntax error (my bad).  I fixed that and it seemed to work even with the apostrophe.  Here's the snippet (note the apostrophe in the comment).

```
if (!tp.file.exists(yearLink)) { // if the yearlink doesn't exist
    await app.vault.createFolder(dir)
}
```
I don't have a great way to roll back to 1.15.0 to determine if it would have failed then.  So, I think it is fixed now.  But, I had eliminated all comments with apostrophes so that templater would work a while ago. I just can't be sure that this example would have failed before.

I'm willing to assume it if fixed and then I'll reopen it if I see it again.  Since it only affects comments, it's not a show stopper.

FYI - templater is very central to what I do and multiple templates get run each day.  It's a great tool.

Scott

PR #821, opened 2022-09-11
fix: strip any comments from being parsed

shabegom:
A naive approach that looks for any lines starting with `//` and removes them before parsing starts.

fixes: #728

### DIFF
--- a/src/core/parser/Parser.ts
+++ b/src/core/parser/Parser.ts
@@ -5,6 +5,7 @@ export class Parser {
         content: string,
         object: Record<string, unknown>
     ): Promise<string> {
+        content = content.replace(/\/\/.*\n/g, "");
         content = (await renderAsync(content, object, {
             varName: "tp",
             parse: {


GitMurf:
@shabegom I know this probably seems like a real corner case but I actually have reasons in some of my current Templater scripts where I would have `//` slashes show up within strings that are valid. So this PR would create an issue here for example when using a template literal.

image

Test code to use:

```js
let myTestString = `    // indented comment EOL
// comment at beginning of line EOL
//comment with no leading space EOL
let dumVar = "foo bar"; // comment at end of line EOL
let falsePositive = \`I am a string template literal with // slashes in it on purpose!\`;
`;
myTestString.replace(/\/\/.*\n/g, "");
```

GitMurf:
I think a "safer" approach would be the following which also accounts for multi line comments `/* my comment */`.

```js
let myTestString = `
let goodLine = "starting";
/* Multi-line
comments over many line
end line of comment */
let normalLine = "URL with slashes https://www.espn.com needs to stay!";
let normalLine2 = "URL with slashes https://www.espn.com needs to stay!"; //but comment needs to leave
    // indented comment EOL
// comment at beginning of line EOL
//comment with no leading space EOL
let numVar = 34; // comment at end of line with number EOL
let dumVar = "foo bar"; // comment at end of line with string double quote EOL
let dumVar = 'foo bar'; // comment at end of line with string single quote EOL
let dumVar = \`foo bar\`; // comment at end of line with string back tick EOL
let dumVar2 = "foo bar" + " more baz" + "this is a good // comment to keep"; // comment at end of line but multiple quote strings before EOL
let dumVar3 = "foo bar" + " more baz" + "this is a good // comment to keep";
let falsePositive = \`I am a string template literal with // slashes in it on purpose!\`;
let falsePositive = \`I am a string template literal with // slashes in it on purpose!\`; // This is a real comment though

let cornerCase = "When a code comment is at end of line with string in front and also comment has quotes in it..."; // the comment has "quotes", or single ' quote or backtick \` in it
`;

// remove single and multi line comments except for one edge corner case
myTestString = myTestString.replace(/(^([^"'`\n]*?)\/\/[^\n]*$|^([^\n]*?)\/\/[^"'`\n]*$|^[\s]*\/\*[\s\S]+?\*\/)/gm, '$2$3');
console.log(myTestString);
```

GitMurf:
@shabegom this should be a solid solution (although please see if you can think of any other corner cases I am not thinking of)...

``myTestString = myTestString.replace(/(^([^"'`\n]*?)\/\/[^\n]*$|^([^\n]*?)\/\/[^"'`\n]*$|^[\s]*\/\*[\s\S]+?\*\/)/gm, '$2$3');``

The regex above is broken into three OR `|` groups explained below:

1) ```^([^"'`\n]*?)\/\/[^\n]*$```
    - Any line starting with `//`
    - Any line starting with spaces or tabs and then `//`
    - Any line with code in front and comment at end of line BUT cannot have any string characters [`"` or `'` or `` ` `` (backtick)] before the comment starts. For example, this does NOT qualify: `let this = "that"; // comment at end` but this WOULD qualify: `let this = 43; // comment at end`.

2) ```^([^\n]*?)\/\/[^"'`\n]*$```
    - This account for lines with code in front and comments at end where there ARE string characters before the comment BUT there are NOT string characters within the comment itself
    - The logic here is that if the comment `//` slashes were actually part of a string then there would be a closing string character [`"` or `'` or `` ` `` (backtick)] within the comment after the `//` slashes
    - This qualifies as a valid match: `let this = "that"; // comment at end looks good`
    - This does NOT qualify as a valid match: `let this = "that"; // comment doesn't work because of apostrophe (single quote)`
    - This also does NOT qualify: ``let falsePositive = `I am a string template literal with // slashes in it on purpose so these slashes do NOT match as a comment because my backtick is after this to close out the template literal!`;``

3) ```^[\s]*\/\*[\s\S]+?\*\/``` - This handles multiline comments starting with `/*` and ending with `*/`.

Let me know if you have any questions.

GitMurf:
Ok here is the last addition I am going to make. Added 3 more scenarios where you can have for example an `'` apostrophe single quote in the comment as long as you do not have a single quote before the comment in the code at the beginning of the line. You can still have `"quotes"` and ``backtick ` `` in front. And then vice versa for the other two... as long as a string character is not before the comment it can be in the comment at end on a mixed line code + comment on same line. See examples below with Pass / Fail to explain.

```js
let myTestString = `
let cornerCase1Pass = "PASS double quote string"; // comment can have apostrophe ' (single quote) OR \` backtick but can NOT have double quote
let cornerCase1Fail = "FAIL double quote string"; // comment canNOT have "double quotes" which this does
let cornerCase2Pass = 'PASS single quote string'; // comment can have " (double quote) OR \` backtick but can NOT have single quotes
let cornerCase2Fail = 'FAIL single quote string'; // comment canNOT have 'single quotes' which this does
let cornerCase3Pass = \`PASS backtick string\`; // comment can have apostrophe ' (single quote) OR "double quotse" but can NOT have backticks
let cornerCase3Fail = \`FAIL backtick string\`; // comment canNOT have \`backticks\` which this does

let apostropheSolution = "now this WORKS because double quotes before comment and single quote after"; // comment doesn't work because of apostrophe (single quote)
`;

myTestString = myTestString.replace(/(^([^"'`\n]*?)\/\/[^\n]*$|^([^\n]*?)\/\/[^"'`\n]*$|^([^\n"']*?)\/\/[^`\n]*$|^([^\n"`]*?)\/\/[^'\n]*$|^([^\n`']*?)\/\/[^"\n]*$|^[\s]*\/\*[\s\S]+?\*\/)/gm, '$2$3$4$5$6');
console.log(myTestString);
```

GitMurf:
Here is the screenshot showing the console logs results of the tests.

image

GitMurf:
To be painfully clear since a lot of commentary above, my proposed solution to your commit would be the following:

```js
content = content.replace(/(^([^"'`\n]*?)\/\/[^\n]*$|^([^\n]*?)\/\/[^"'`\n]*$|^([^\n"']*?)\/\/[^`\n]*$|^([^\n"`]*?)\/\/[^'\n]*$|^([^\n`']*?)\/\/[^"\n]*$|^[\s]*\/\*[\s\S]+?\*\/)/gm, "$2$3$4$5$6");
```

shabegom:
I appreciate the effort, but that's way too complex an brittle to use. 

The original issue is that apostrophes sin comment caused the commands to fail. Perhaps the easier approach would be to escape the apostrophes.

If you would like to open your own PR for this I can review. Closing this one.

GitMurf:
And just to be clear, the exception above with inline comments would NOT add any regression. It would just operate like it does today with the apostrophe issue. But we would at least eliminate this apostrophe issue for regular comments and `/* multi line comments */`.

`let foo = "bar"; // this is an inline comment and would stay status quo like today`

Thoughts?

shabegom:
@GitMurf Let's ignore the comments aspect for a moment. All we really need to do is escape the apostrophe. 

From this: `what's going on`
To this: `what\'s going on`

I just ran a test and escaped apostrophes work just fine AND if you have an escaped apostrophe in the output string the escape character gets stripped. 

So all we need to do is find and apostrophe character between two letters and put a backslash in front of it.

GitMurf:
@shabegom sounds good. And we always have my quantum physics regex statement saved in this closed PR in case we ever decide it is needed to re-address :-) 🤣 

Do you have this new apostrophe escape code anywhere for me to review and make sure I don't see any corner cases unaccounted for? Happy to review :)

PR #851, opened 2022-10-03
Fix tp.file.exists() example

ilandikov:
tp.file.exists() is async, but the docs are not reflecting that. Also it requires full path as an argument to avoid collisions of files with same name in different folders


shabegom (approved):
LGTM

Issue #873, opened 2022-10-18
Obsidian freezes when a large file is received via sync or dropped to the file explorer from the file system


zsviczian:
I am raising this issue again, since the original was never addressed, however it continues to cause problems :( See also #439 

Please take a look at this issue raised for Excalidraw [#805](https://github.com/zsviczian/obsidian-excalidraw-plugin/issues/805#issuecomment-1281475586) recently. In case of this example, the drawing is 2.8MB large. If I drop (or sync for the first time) the file into Obsidian, Obsidian freezes. Maybe it is just slow processing, however, I killed Obsidian after 10 minutes of waiting, so I don't know. Please test using this file: [math.md](https://github.com/zsviczian/obsidian-excalidraw-plugin/files/9804455/math.md)

If I disable Templater before dropping the new file the issue does not happen.

I could disable "Trigger Templater on new file creation" however, this is not desirable since I use this function for automation such as my daily notes template.

I recommend adding an option to disable trigger on new file creation for files larger than a user set size limit. I would set the default size limit to 100Kb or so.

Thank you!

shabegom:
Duplicate of #315

zsviczian:
Doing some debugging, this is where the process gets stuck:

```typescript
var parseOpenReg = new RegExp("([^]*?)" + escapeRegExp(config2.tags[0]) + "(-|_)?\\s*(" + prefixes + ")?\\s*(?![\\s+\\-_" + prefixes + "])", "g");
var parseCloseReg = new RegExp("'|\"|`|\\/\\*|(\\s*(-|_)?" + escapeRegExp(config2.tags[1]) + ")", "g");
var m;
while (m = parseOpenReg.exec(str)) {
  lastIndex = m[0].length + m.index;
  var precedingString = m[1];
  var wsLeft = m[2];
  var prefix = m[3] || "";
```

More precisely the `parseOpenReg.exec(str)` the regexp parsing seems to take forever. In case of math.md referenced above `str.length === 2945580`.

As a quick fix I would put in a size limit check here

https://github.com/SilentVoid13/Templater/blob/0874f2491b8e714856e6d308f8fc3594cb88111f/src/core/Templater.ts#L421-L423

zsviczian:
I created #874 with a hard coded file size limit of 10000 to skip parsing of large files.
10000 is proposed based on a quick analysis of my vault:
[...]

PR #874, opened 2022-10-18
fix: parsing freezes on large files

zsviczian:
A hard coded fix for #873.

Based on a quick analysis of my vault, I propose to hard coded file size limit of 10000 to skip the on file create parsing of large files.

I have 5675 markdown files in my vault. Over 90% percent of these have a size below the proposed 10000 size limit.

### DIFF
--- a/src/core/Templater.ts
+++ b/src/core/Templater.ts
@@ -419,7 +419,11 @@ export class Templater {
             }
             await templater.write_template_to_file(template_file, file);
         } else {
-            await templater.overwrite_file_commands(file);
+            if(file.stat.size <= 10000) { //https://github.com/SilentVoid13/Templater/issues/873
+                await templater.overwrite_file_commands(file);
+            } else {
+                console.log(`Templater skipped parsing ${file.path} because file size exceeds 10000`);
+            }
         }
     }
 


zsviczian:
Hi @shabegom, please take a look at this. 

This would be an important fix in my view. Without it, I need to keep `Trigger Templater on new File creation` turned off, and only turn it on before creating my DNP (or other relevant templates). Since I've turned the `trigger` off, Obsidian sync has been running smooth again on all my devices. Before turning it off, I had to kill Obsidian several times before it was able to synchronize everything. I expect that other Excalidraw users also use Templater and Obsidian Sync and then face the same hurdles every day.

shabegom:
Yep. Will merge when I have the time.

src/core/Templater.ts
```diff
@@ -419,7 +419,11 @@ export class Templater {
             }
             await templater.write_template_to_file(template_file, file);
         } else {
-            await templater.overwrite_file_commands(file);
+            if(file.stat.size <= 10000) { //https://github.com/SilentVoid13/Templater/issues/873
```
shabegom, line 422: would be better if this check was for the reverse with a `return` like the expressions above.

src/core/Templater.ts
```diff
@@ -419,7 +419,11 @@ export class Templater {
...
+                await templater.overwrite_file_commands(file);
+            } else {
+                console.log(`Templater skipped parsing ${file.path} because file size exceeds 10000`);
```
shabegom, line 425: would be better if this used the errorWrapper function

shabegom (approved):

Issue #878, opened 2022-10-23
Bug: tp.file.exists in Obsidian v1.0.0 always returns false for files; 


jlharri:
Obsidian version 1.0.0, installer build v0.15.8
Templater version 0.11.13

I've tried multiple tests and tp.file.exists returns false for all files. It will return true for a folder. 

Here's a test template I created and its output. Note that I've tried other hard-coded file names. I thought using tp.file.title might be the most obvious example of a failure. 

```
<%*
console.log("--------");
console.log("Testing Templater tp.file.exists");
console.log("Obsidian version 1.0.0 - installer version v0.15.8");
console.log("Templater Version "+app.internalPlugins.app.plugins.manifests["templater-obsidian"].minAppVersion);
console.log("");

fn = tp.file.title;
console.log("tp.file.title <"+fn+"> exists? "+(await tp.file.exists(fn)));
fn = tp.file.folder(false);
console.log("tp.file.folder <"+fn+"> exists? "+(await tp.file.exists(fn)));
fn = "bogus file name";
console.log("Bogus file <"+fn+"> exists? "+(await tp.file.exists(fn)));
%>
```

Output copied from JS console: 
```
Testing Templater tp.file.exists
Obsidian version 1.0.0 - installer version v0.15.8
Templater Version 0.11.13

tp.file.title <Test Templater Doc> exists? false
tp.file.folder <Test> exists? true
Bogus file <bogus file name> exists? false
```

btw, great plugin! It'd be nice to have the file.exists test part of tp.file.create_new as an optional parameter create_if_new. 

Jerry

AB1908:
You need to `await` the `exist()` I think.

jlharri:
I do...it's in the code snippet above. 

I figured it out. The documentation isn't clear about whether I can pass in just that name of a file, the name+extension, or the fully qualified path relative to the Vault. It's the last one. For example, 

For a file located at Test/Templater Tests/Test Doc.md :
- tp.file.exists("Test Doc") => false
- tp.file.exists("Test Doc.md") => false
- tp.file.exists("Test/Templater Tests/Test Doc.md") => true

So I put this into my template and it works: 
```
fn = tp.file.folder(true)+"/"+tp.file.title+".md";
console.log("tp.file.title <"+fn+"> exists? "+(await tp.file.exists(fn)));
```
I hate that I have to hard-code the extension, but I realize most files in Obsidian are markdown. 

I suggest updating the documentation to be clearer about what is needed.

[...]

PR #879, opened 2022-10-23
Bug: tp.file.exists in Obsidian v1.0.0 always returns false for files;

jlharri:
Addresses #878 in 2 ways:
1) Fixes documentation to reflect that tp.file.exists requires a full path. 
2) Adds a call to normalizePath() inside exists() function so that templates can work across Windows and *nix OS'.

### DIFF
--- a/docs/documentation.toml
+++ b/docs/documentation.toml
@@ -161,7 +161,7 @@ description = "The content to append after the active cursor"
 
 [tp.file.functions.exists]
 name = "exists"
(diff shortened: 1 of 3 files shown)


jlharri:
Let me know if I've screwed something up with the PR. I had to make a second commit for InternalModuleFile.ts.

shabegom:
Thanks for the PR. Will review when I have some time. At a glance it looks good!

PR #885, opened 2022-10-27
Set activeFile to newly created note

ys:
This allows plugins using app.workspace.getActiveFile() to be called … from template

I am using a plugin that I call via the command 
```
<% app.workspace.executeCommandById("lol") %>
```

and that plugin uses the current_file, I have no power over that and wanted this to work like for new daily notes made via the core plugin.

This allows the `Create new note from template` to behave like the new daily note.

### DIFF
--- a/src/core/Templater.ts
+++ b/src/core/Templater.ts
@@ -124,7 +124,10 @@ export class Templater {
             folder,
             filename ?? "Untitled"
         );
-
+        
+        // Set activeFile to newly created note
+        await app.workspace.getLeaf(false).openFile(created_note, { state: { mode: "source" }, })
+        
         let running_config: RunningConfig;
         let output_content: string;
         if (template instanceof TFile) {


src/core/Templater.ts
```diff
@@ -124,7 +124,10 @@ export class Templater {
             folder,
             filename ?? "Untitled"
         );
-
+        
+        // Set activeFile to newly created note
+        await app.workspace.getLeaf(false).openFile(created_note, { state: { mode: "source" }, })
```
shabegom, line 129: This is happening on line 170 below if `open_new_note` is true. What is the purpose of adding it here?

shabegom (commented):
Thanks for the PR! I don't understand why the same logic on line 170 isn't setting the `activeFile`.

src/core/Templater.ts
ys: That allows calls in the template to use the currentFile if they do. Some plugins assert that they function on current file. I am using the Shortcut Launcher to edit some text with data I cannot get from obsidian directly and if I don't have this call it will affect the previous currentFile

ys: running the template is the part where I like having the currentfile set to what the template is.

Does that make sense? Daily notes using templater kind of does that

shabegom: Ok I understand, but should the logic that sets the current file be removed below? There is also a check in place if the file should be opened.

I want to make sure adding this won't break anyone's existing workflow.

ys: I guess the logic below could be partly removed. Let me review the code more.

For other workflows... Hard to tell

Zachatoo:
Doing some cleanup, feel free to reopen if you'd still like to contribute!

Issue #884, opened 2022-10-27
Feature Request: Jumping (`tp.file.cursor(n)`) should also automatically enter insert mode


chrisgrieser:
A small convenience feature thing, but would be nice?

shabegom:
released!

PR #886, opened 2022-10-28
Jumping cursor loc. enters insert mode (#884)

chrisgrieser:
implements FR #884.

tested locally on my device and seems to work fine.

### DIFF
--- a/src/editor/CursorJumper.ts
+++ b/src/editor/CursorJumper.ts
@@ -25,6 +25,15 @@ export class CursorJumper {
             await app.vault.modify(active_file, new_content as string);
             this.set_cursor_location(positions);
         }
+
+        // enter insert mode for vim users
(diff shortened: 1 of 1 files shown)


src/editor/CursorJumper.ts
```diff
@@ -25,6 +25,15 @@ export class CursorJumper {
             await app.vault.modify(active_file, new_content as string);
             this.set_cursor_location(positions);
         }
+
+        // enter insert mode for vim users
+        if (app.vault.getConfig("vimMode")) {
+           // @ts-ignore
```
shabegom, line 31: @chrisgrieser I'd prefer if the "obsidian" module was extended so we avoid using @ts-ignore

shabegom (approved):
I don't like using @ts-ignore

PR #889, opened 2022-10-29
feat: enable syntax highlighting in cm6

### LINKED ISSUE #491: Syntax Highlighting not working in CM6

-   OS: macOS 12.0.1
-   Templater version: 1.9.9
-   Obsidian version: Obsidian 0.13.14
-   Templater settings: Syntax Highlighting enabled

**Describe the bug**
Syntax Highlighting does not work with the new editor (CM6). When using the Legacy Editor, Syntax Highlighting works again.

**Screenshots**
Screenshot 2021-12-30 15 46 38

### LINKED ISSUE #854: Conflict between Templater и Smart-Typography
One of the features of the [Smart-Typography](https://github.com/mgmeyers/obsidian-smart-typography) plugin is automatically turning standard double quotes into curly double quotes ("" → ““). This is a very subtle change that is hard to notice, and I didn't even know about this until I discovered my issue (I initially downloaded the plugin just for em-dashes).

At some point I started setting up the plugin Templater and encountered an issue where, when creating simple templates (for example, just inserting the creation date `<% tp.file.creation_date(“D MMMM YYYY“)%>`), while processing the template, Templater crashed with a parsing error 
```
Templater Error: Template parsing error, aborting. 
 Bad template syntax
```
and an error message in the console that is not interpretable
```
Invalid or unexpected token
===========================
var tR='',__l,__lP,include=E.include.bind(E),includeFile=E.includeFile.bind(E)
function layout(p,d){__l=p;__lP=d}
const _prs = [];
_prs.push(tp.file.creation_date(“D MMMM YYYY“));
tR+='rJ2KqXzxQg'
tR+='\n\n'
const _rst = await Promise.all(_prs);
tR = tR.replace(/rJ2KqXzxQg/g, () => _rst.shift());
if(__l)tR=await includeFile(__l,Object.assign(tp,{body:tR},__lP))
if(cb){cb(null,tR)} return tR
```

After some digging around, I finally discovered that the issue was in Templater not being able to parse the curly quotes, and everything was, of course, fixed by using the standard quotes.

## Preventive measures?
1. The error message is very hard to interpret. At least it should be able to better pinpoint the lines where the error lies? Instead of giving lots of code that I don't recognize at all and get no useful info from.
2. Maybe something can be done specifically against cases like this? For example, inform the user their Templater code blocks include unparseable characters.

mgmeyers:
Closes #854; Closes #491

cc: @shabegom

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
     "dependencies": {
         "@codemirror/state": "^6.1.1",
         "@codemirror/view": "^6.2.3",
+        "@codemirror/language": "https://github.com/lishid/cm-language",
         "@popperjs/core": "^2.11.2",
         "child_process": "^1.0.2",
         "eta": "github:SilentVoid13/eta#custom-eta"
--- a/src/editor/Editor.ts
+++ b/src/editor/Editor.ts
@@ -8,6 +8,7 @@ import { Autocomplete } from "editor/Autocomplete";
 
 import "editor/mode/javascript";
 import "editor/mode/custom_overlay";
+import { StreamLanguage } from "@codemirror/language";
 //import "editor/mode/show-hint";
 
 const TP_CMD_TOKEN_CLASS = "templater-command";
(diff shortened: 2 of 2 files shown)


shabegom:
released! Thank you!!!

PR #894, opened 2022-10-29
Added content to help files. Await on tp.system cmds

Negative-light:
Currently, the Execution commands cannot save the output of a suggester or prompt to a variable w/o an await. Added an example and some content about the change.

### DIFF
--- a/docs/src/commands/execution-command.md
+++ b/docs/src/commands/execution-command.md
@@ -24,7 +24,7 @@ For example, the following command: `<%* tR += "test" %>` will output `test`.
 
 ### Suggesters and Prompts
 
-It is important to note that the `tp.system.prompt()` and `tp.system.suggester()` both requrire a `await` statment to save the value to a variable
+It is important to note that the `tp.system.prompt()` and `tp.system.suggester()` both require a `await` statment to save the value to a variable
 
 ## Examples
 
--- a/docs/src/internal-functions/internal-modules/system-module.md
+++ b/docs/src/internal-functions/internal-modules/system-module.md
@@ -43,7 +43,7 @@ Picked file: [[<% (await tp.system.suggester((item) => item.basename, app.vault.
 
 
 <%*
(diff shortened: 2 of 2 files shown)


docs/src/internal-functions/internal-modules/system-module.md
```diff
@@ -40,4 +40,11 @@ Mood today: <% tp.system.prompt("What is your mood today ?", "happy") %>
 
 Mood today: <% tp.system.suggester(["Happy", "Sad", "Confused"], ["Happy", "Sad", "Confused"]) %>
 Picked file: [[<% (await tp.system.suggester((item) => item.basename, app.vault.getMarkdownFiles())).basename %>]]
-```
\ No newline at end of file
+
+
+<%*
+const execution_value = tp.system.suggester(["Yes", "No"], ["true", "false"])
```
AB1908, line 46: You're missing an await here

Negative-light: That would be problematic. Fixing **ASAP**

Negative-light: Ok this is fixed.

docs/src/commands/execution-command.md
```diff
@@ -22,6 +22,10 @@ This means that, to output something from a JS execution command, you just need
 
 For example, the following command: `<%* tR += "test" %>` will output `test`.
 
+### Suggesters and Prompts
+
+It is important to note that the `tp.system.prompt()` and `tp.system.suggester()` both requrire a `await` statment to save the value to a variable
```
shabegom, line 27: spelling error. should be "require"

shabegom (changes requested):
One spelling error, otherwise LGTM

docs/src/commands/execution-command.md
Negative-light: Should be updated, sorry about that.

PR #903, opened 2022-11-02
Improve Syntax Highlighting

### LINKED ISSUE #901: Templater should use the user colours for syntax highlighting

Obsidian 1.0 introduced a new simple system for customising the majority of colours in the application with simple variables. Templater doesn't respect the user's colour settings, instead using its own colours for syntax highlighting.

Templater should utilise the user-defined `--code-...` style variables for its syntax highlighting.

**Describe alternatives you've considered** Overriding things manually.

**Additional context**

iamrecursion:
This PR aims to add some improvements to how syntax highlighting is handled for users of Templater.

Firstly, it adds a separate toggle that allows the user to enable syntax highlighting for mobile independently of for desktop. This allows users that don't use Live Preview (where highlighting is currently broken) on mobile to have the highlighting.

Secondly, it changes the styling used by Templater's syntax highlighting to use user-defined colours under the Obsidian 1.0 style sheet. It aims to match the built-in JavaScript syntax highlighting as closely as possible.

It is backwards compatible in that highlighting will continue to work for earlier versions of obsidian. However, some classes have been consolidated to provide a more coherent style experience, so some colours may change.

Fixes #901


SilentVoid13:
Hey @iamrecursion thanks a lot for the PR! I suck in CSS and hacked a thing that was more or less working so I kept it at that.
Quick question: you set the color of tags to a custom color

```css
.cm-s-obsidian .cm-templater-command.cm-templater-interpolation-tag {
    color: var(--code-property, #008bff);
}

.cm-s-obsidian .cm-templater-command.cm-templater-execution-tag {
    color: var(--code-function, #c0d700);
}
```

Do you think that should be customizable?

iamrecursion:
My thought was that it makes most sense for those to match the user theme as well. They stick out quite badly otherwise. I'm not dead set on that change however.

SilentVoid13:
Yeah I guess that's okay too, not sure what's the best either. Thanks again for your contribution

iamrecursion:
An absolute pleasure!

PR #916, opened 2022-11-11
tp.config.template_file is mentioned as tp.file

ilandikov:
The function is in the tp.**config** module, but the documentation mentions as tp.**file**.


ilandikov:
@SilentVoid13 could you pls  help me with this PR? I'm not a GH expert =) just trying to contribute to an awesome plugin

PR #925, opened 2022-11-17
Prevent inserting invalid ctime

### LINKED ISSUE #876: Templater tp.file.creation date producing "1 January 1970" when creation date is not available
### Steps to reproduce

- Create a file in the vault with an unaccesible creation date. I got such a file whlie creating a text file on my HDD https://gitlab.gnome.org/GNOME/nautilus/-/issues/1734. Don't know how to intentionally create such a file. In a different environment
- Create a simple template
```
<% tp.file.creation_date("D MMMM YYYY") %>
```
- In the file without a creation date, insert this template

### Expected result
- Due to the creation date of the file not being available, the `tp.file.creation_date` should be replaced with an "ERROR" of some sort
- A warning should pop up informing that the creation date is not available for some reason

### Actual result
The date is replaced with "1 January 1970"

### Environment

- Operating system: Linux Mint

Zachatoo:
Fixes #876

### DIFF
--- a/src/core/functions/internal_functions/file/InternalModuleFile.ts
+++ b/src/core/functions/internal_functions/file/InternalModuleFile.ts
@@ -94,6 +94,9 @@ export class InternalModuleFile extends InternalModule {
 
     generate_creation_date(): (format?: string) => string {
         return (format = "YYYY-MM-DD HH:mm") => {
+            if (!this.config.target_file.stat.ctime) {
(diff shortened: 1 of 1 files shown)


AB1908 (approved):

Zachatoo:
Closing because I don't think this is the correct solution. A warning of some kind would be sufficient, don't prevent template execution.

PR #947, opened 2022-12-07
Preserve order of function arguments in documentation

TomHall2020:
Encoding function arguments as an array of tables instead of a table of tables in `documentation.toml` allows us to iterate over them in the order declared instead of being sorted alphabetically in the docs. 
Works correctly in my local build of the docs.

### DIFF
--- a/docs/documentation.toml
+++ b/docs/documentation.toml
@@ -37,19 +37,19 @@ name = "now"
 description = "Retrieves the date."
 definition = "tp.date.now(format: string = \"YYYY-MM-DD\", offset?: number⎮string, reference?: string, reference_format?: string)"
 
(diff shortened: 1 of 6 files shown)


TomHall2020:
Dumped the output onto netlify: https://jocular-lokum-1cb415.netlify.app/

Zachatoo:
I was JUST looking at the docs earlier today and noticed the arguments out of order, this is awesome!

AB1908:
`tp.file.move` looks like it's missing an arg but the rest looks good.

TomHall2020:
> `tp.file.move` looks like it's missing an arg but the rest looks good.

Checked on the live current version and its missing the same argument there too so must have just been missing off the docs.

PR #957, opened 2022-12-15
Allow Trigger Create new note from Insert Template command

### LINKED ISSUE #425: Cannot create a new note using Template Hotkeys (configuration allowed by templater to bind to hot keys)

-   Windows 10
-   Templater version: 1.9.9
-   Obsidian version: 0.12.19

**Describe the bug**
NOT sure if this is a bug or just me doing something dumb.  
I cannot create a new note using Template Hotkeys. 
I created a couple of Template Hotkeys in Templater settings where I bound some specific templates to a hotkey. When I open Obsidian and no file is open, I try creating a new note using the Template hotkeys. But nothing is created and I get the error "no active view. can't append templates".  On the other hand, if I already have a note open, then the hotkey would work.

**Expected behavior**
A new note should be created with the template when pressing the Template Hotkeys even if no file is open.

mnaoumov:
Fixes #425

### DIFF
--- a/src/core/Templater.ts
+++ b/src/core/Templater.ts
@@ -188,9 +188,14 @@ export class Templater {
     async append_template_to_active_file(template_file: TFile): Promise<void> {
         const active_view = app.workspace.getActiveViewOfType(MarkdownView);
         if (active_view === null) {
-            log_error(
-                new TemplaterError("No active view, can't append templates.")
-            );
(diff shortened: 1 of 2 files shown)


Zachatoo:
Cleaning up stale PRs.

mnaoumov:
@Zachatoo Can I do something for it to be reopen and merged?

Issue #471, opened 2021-12-13
allow folder templates to match on file name


229c9cf0:
**Problem**
It's possible to apply per-folder templates, but there are often different styles of files combined in a folder that differ by file name pattern. Some examples:

 *  logs / journals can have monthly, weekly, and daily notes, so the patterns might be e.g.

     + Log/`YYYY`/`YYYY`-`MM`
     + Log/`YYYY`/`YYYY`-W`WW`
     + Log/`YYYY`/`MM`/`YYYY`-`MM`-`DD`

    The current capabilities would only allow a folder template for `Log`, which clearly isn't specific enough. Even folder patterns as suggested in #418 would not solve this (since monthly and weekly notes are in the same directory), but being able to match the combination of folder and file pattern as e.g. folder `/Log` and file pattern `????-??`, `????-W??`, or `????-??-??` would fully identify the type (monthly / weekly / daily note.)
 *  In projects, data collections, etc., beside the actual data that should match the folder-specific template, there's often also metadata that should match global metadata templates. `README` files are probably the most common of those, I also have things like `INBOX`, `TODO`, `STATS`, ... Ideally, they they all have exactly the same structure, no matter where in the tree they are. So here the folder pattern would just be `/` and the file patterns `INBOX` / `README` / ..., i.e. no matter where in the tree, this template should be used for this particular file.

**Suggested solution**
Extend the folder templates list from `(folder, template)` to `(folder_pattern, file_pattern?, template)` (this would subsume #418).

If present, a `file_pattern` match should be stronger than even a deep `folder_pattern` match, to allow global structure across all folders. (I think the end result is two layers: (1) both file + folder pattern match, pick the deepest of these. (2) no file pattern and only the folder pattern matches, pick the deepest as done currently.)

Specific flavour of pattern doesn't matter. Glob-style would be nice (no need to escape `/` in paths), but plain regex works too. (Glob implementations exist - e.g. 'minimatch' - but I can't assess whether they're any good.)

As a first step, even allowing only exact matches on file names (no patterns at all) would already solve many problems.

**Alternatives?**
Not aware of any, apart from manual workarounds.

**Additional context**
n/a

shabegom:
Your folder template could match on filename and then use `tp.file.include` to bring in the appropriate template.

harmtemolder:
I ran into this same issue today, so decided to try if using regexes on file paths was feasible. It was. This works:

[...]

PR #974, opened 2022-12-29
Add option to use regex on file path instead of folder templates

harmtemolder:
As discussed in #471 the current folder templates do not cover all use cases. Having the option to use regex on the file path of a new file instead should give advanced users the flexibility they seek.

I've kept the folder templates option intact, letting a user pick between the two. Their toggles are linked, so enabling one will disable the other.

I've also updated the docs to match the current layout of the settings window.

And I made two minor changes to `.gitignore`:
- Exclude `package-lock.json`, because I don't think we need to lock versions, do we?
- Include `.vscode/settings.json`, to allow for synching of workspace settings between our VSCodes, specifically `editor.defaultFormatter`, so that all of use Prettier (and thus the included `.prettierrc.json`)


harmtemolder:
n=1 of course, but I've been using these changes without any hick-ups since December

quanru:
Is it possible to merge with the main branch?

harmtemolder:
> Is it possible to merge with the main branch?

We would need @SilentVoid13 for that

AB1908:
> Exclude package.json, because I don't think we need to lock versions, do we?

We still actually. It's still useful to pin APIs to specific obsidian versions. There's also scripts and dependencies in there that we'd like to start updated. Finally I think Obsidian itself needs the version info in it to install properly.

quanru:
Could you remove the changes to .gitignore and vscode configuration, or have a separate MR, so that this feature can be merged faster? @harmtemolder

harmtemolder:
Sorry about that. I see that I wrote `package.json`, but I actually excluded `package-lock.json`, not `package.json`. I'll remove `.vscode/settings.json`

SINC-G:
Excuse me, Does the PR have made any new progress?

harmtemolder:
The PR was finished a year ago, but I see there are now conflicts because of changes made to the plugin in the time since then. I (or someone else) would need to resolve those before @SilentVoid13 would be able to merge the PR.

@SilentVoid13 Are you interested in merging this PR? If so, I'll take the time to resolve the conflicts

Zachatoo:
I am interested in seeing this functionality added. If you resolve the conflicts and remove the unrelated changes (such as updating the `.gitignore` file) then I will be happy to pull this down to test and merge it if it works on my end.

Ideally, we'd also add unit tests for this, but I don't think it's strictly required, as long as the existing tests still pass.

harmtemolder:
@Zachatoo I've solved all conflicts, built and tested with the file regex templates I use in my vault. Those work as expected. Could you explicitly test that this does not break any existing folder templates somebody might be using?

I've never written unit tests in TypeScript, but adding those for this functionality sounds like a good idea. Let me know how I can help with that.

Zachatoo (approved):
@harmtemolder Thanks for making those changes! PR looks a lot cleaner. Automated tests are passing and existing folder templates functionality appears to be working still.

There were some changes from #1468 that will result in a UX conflict, but I can resolve that after I merge this before the release.

PR #1026, opened 2023-02-07
Add url_only option to random_picture to improve plugin compability

JBrandonS:
This patch adds a url_only option to `tf.web.random_picture()` to enable better compatibility with other plugins, i.e. Banners. This value defaults to `false` to maintain the current behavior. Documentation has been updated to add this option.

### DIFF
--- a/docs/documentation.toml
+++ b/docs/documentation.toml
@@ -324,7 +324,7 @@ definition = "tp.web.daily_quote()"
 [tp.web.functions.random_picture]
 name = "random_picture"
 description = "Gets a random image from https://unsplash.com/"
-definition = "tp.web.random_picture(size?: string, query?: string, include_size?: boolean, raw_output?: boolean)"
+definition = "tp.web.random_picture(size?: string, query?: string, include_size?: boolean, url_only?: boolean)"
 
 [tp.web.functions.random_picture.args.size]
 name = "size"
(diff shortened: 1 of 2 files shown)


slambert:
I was just looking for this! Would love to see it merged.

JBrandonS:
@slambert 

In the mean time, what I have done is save the following script and enabled it in templater settings. This allows you to use the function and not have to use a custom version.

 I wish I had kept attribution for the real author of this code but I found it somewhere online.
```javascript
// scripts/random_banner.js
async function random_banner(size, query) {
	const resp = await fetch(`https://source.unsplash.com/random/${size ?? ""}?${query ?? ""}`);
	if (!resp.ok) {
		return "Error: Couldn't fetch banner";
	}
	return resp.url;
}

module.exports = random_banner;
```

Example usage:

```markdown
---
cssclass: note
aliases:
tags:
banner: "<% tp.user.random_banner("1080x800", "banner," + tp.file.title) %>"
---
```

slambert:
Being sort of tenacious I kept going and trying to solve it myself without the code.

I spent some time and figured out the unsplash URL structure. A lot of stuff I found about the URL structure is out of date and no longer works. 

Here's what works best:

`https://source.unsplash.com/random?sig=UNIQUESTRING/&SEARCHTERM`

So

https://source.unsplash.com/random?sig=1/&birthday
https://source.unsplash.com/random?sig=2/&birthday

would give you this:

 


But I didn't want repeat images if I had the same search term, so I needed that `sig=` value to be more complex that just `1` or `2`.

So for a daily note, I used the name of the note as the string. I just change the search keyword every so often depending on my mood:
`banner: "https://source.unsplash.com/random?sig=<% tp.file.title.slice(0,10) %>/&abstract"`

And for other files, I used the file creation date for the string. I also added a prompt so I could enter in a keyword for the image.

`banner: "https://source.unsplash.com/random?<% tp.file.creation_date("YYYYmmDDHHmm") %>/&<% tp.system.prompt('What photo?', 'abstract', false, false) %>"`

**And it works!**

~~I didn't bother trying~~ I tried to incorporate a size specification into the URL because I am tenacious. But couldn't get anything to work. Again, everything I found in a search was out of date or didn't work.

Zachatoo:
Going to close this in favor of #1385, since I believe it fulfills the same use case, thanks!

PR #1078, opened 2023-04-05
Add option to show relative paths instead of only filenames

DraggonFantasy:
Personally I have templates grouped by categories in different folders and it's inconvenient to select the template by looking only on its filename. So I added an option to include the subfolders in the templates list


Zachatoo:
I think I'd prefer to always show the path relative to the template folder path, without a toggle. That feels very "Obsidian"y, that's how the quick switcher works.

Zachatoo:
Closing in favor of #1418, thank you for your initial work on this!

PR #1139, opened 2023-07-08
Add optional initial_query to suggester

### LINKED ISSUE #555: Default value for suggester
I want to have a default value for suggester. Placeholder is ok, but having a default value is often helpful, e.g., when I want to search all markdown files based on a previous user input. 

Add a param for default value

Using placeholder, and have user type in the input again.

mnaoumov:
Fixes #555

### DIFF
--- a/docs/documentation.toml
+++ b/docs/documentation.toml
@@ -555,7 +555,7 @@ selected value: <% value %>"""
 [tp.system.functions.suggester]
 name = "suggester"
 description = "Spawns a suggester prompt and returns the user's chosen item."
-definition = "tp.system.suggester(text_items: string[] | ((item: T) => string), items: T[], throw_on_cancel: boolean = false, placeholder: string = \"\", limit?: number = undefined, default_value?: string = undefined)"
+definition = "tp.system.suggester(text_items: string[] | ((item: T) => string), items: T[], throw_on_cancel: boolean = false, placeholder: string = \"\", limit?: number = undefined, default_value?: T = undefined)"
 
 [[tp.system.functions.suggester.args]]
 name = "text_items"
--- a/docs/documentation.toml
+++ b/docs/documentation.toml
@@ -666,7 +666,7 @@ let selectedValues = await tp.system.multi_suggester(["Happy", "Sad", "Confused"
 # <% selectedValues %>
 selected values: <% selectedValues %>"""
 
-[[tp.system.functions.suggester.examples]]
+[[tp.system.functions.multi_suggester.examples]]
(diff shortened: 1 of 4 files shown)


Zachatoo:
Could you also update the docs as part of this PR?

AB1908:
Can someone remind me how this is different from placeholder? Also, for conformity, should we rename it to `default_value`?

tektrip-biggles:
> Can someone remind me how this is different from placeholder? Also, for conformity, should we rename it to `default_value`?

My understanding is that `placeholder` simply appears in the prompt box as an aide memoire, whereas `default_value` would be a valid option from the list presented to the user. E.g. I want to be prompted for task priority I might use:

```
<% tp.system.suggester(
  (item) => item, 
  ["1: Highest", "2: High", "3: Medium", "4: Low", "6: Lowest"], 
  false, 
  "Priority",
   5, 
  "3: Medium") 
%>
```

This way if I press enter as soon as the prompt appears, I'll select "3: Medium" as priority instead of "1: Highest" as currently. This saves two keypresses in the average case and four keypresses in the worst case, which can add up quickly when, say, adding a ton of task notes.

mnaoumov:
@Zachatoo ready to review

docs/documentation.toml
```diff
@@ -650,6 +666,14 @@ let selectedValues = await tp.system.multi_suggester(["Happy", "Sad", "Confused"
 # <% selectedValues %>
 selected values: <% selectedValues %>"""
 
+[[tp.system.functions.suggester.examples]]
```
Zachatoo:
```suggestion
[[tp.system.functions.multi_suggester.examples]]
```

src/core/functions/internal_functions/system/MultiSuggesterModal.ts
```diff
@@ -135,10 +137,14 @@ class Suggester<T> extends AbstractInputSuggest<T> {
...
         private getItemText: (item: T) => string,
         private items: T[],
-        limit?: number
+        limit?: number,
+        default_value?: string
     ) {
         super(app, textInputEl);
         limit && (this.limit = limit);
+        if (default_value) {
+            this.setValue(default_value);
```
Zachatoo: This doesn't appear to be working as expected, the user still has to select the option for it to work for the multi-suggester.

IMO we should pass an array of options in the template as the default values and have them pre-selected, instead of just setting the input value.

mnaoumov:
@Zachatoo ready for review

Zachatoo (approved):
Great, thanks!

PR #1149, opened 2023-07-23
Add `multiselect`  function to `system` module

sujyotraut:
https://github.com/SilentVoid13/Templater/assets/70472979/c9fba4e8-8d39-40ef-9fb1-5d7e79342dfe

https://github.com/SilentVoid13/Templater/assets/70472979/2cf56311-b553-4c7a-bf80-b699b6adbc04

Is it only happening in my system or same happens in others as well.
Please tell me if it is same on your system as well, and provide feedback on how to solve it if possible.

### DIFF
--- a/docs/documentation.toml
+++ b/docs/documentation.toml
@@ -287,6 +287,39 @@ description = "Throws an error if the prompt is canceled, instead of returning a
 name = "multiline"
 description = "If set to true, the input field will be a multiline textarea"
 
+[tp.system.functions.multiselect]
+name = "multiselect"
+description = "Spawns a multiselect prompt and returns the user's chosen items."
+definition = "tp.system.multiselect(text_items: string[] ⎮ ((item: T) => string), items: T[], unrestricted: boolean = false, throw_on_cancel: boolean = false, placeholder: string = \"\", title: string = \"Multiselect\", limit?: number = undefined)"
+
+[tp.system.functions.multiselect.args.text_items]
+name = "text_items"
+description = "Array of strings representing the text that will be displayed for each item in the multiselect prompt. This can also be a function that maps an item to its text representation."
+
+[tp.system.functions.multiselect.args.items]
+name = "items"
+description = "Array containing the values of each item in the correct order."
+
+[tp.system.functions.multiselect.args.unrestricted]
+name = "unrestricted"
+description = "A boolean which represents whether the values outside suggestions can be selected or not"
+
+[tp.system.functions.multiselect.args.throw_on_cancel]
+name = "throw_on_cancel"
+description = "Throws an error if the prompt is canceled, instead of returning a `null` value"
+
+[tp.system.functions.multiselect.args.placeholder]
+name = "placeholder"
+description = "Placeholder string of the prompt"
+
+[tp.system.functions.multiselect.args.title]
+name = "title"
+description = "Title of multiselect prompt"
+
+[tp.system.functions.multiselect.args.limit]
+name = "limit"
+description = "Limit the number of items rendered at once (useful to improve performance when displaying large lists)"
+
 [tp.system.functions.suggester]
 name = "suggester"
 description = "Spawns a suggester prompt and returns the user's chosen item."
--- a/src/core/functions/internal_functions/system/MultiSelectModal.ts
+++ b/src/core/functions/internal_functions/system/MultiSelectModal.ts
@@ -1,0 +1,229 @@
+import { Modal, TextComponent, fuzzySearch, prepareQuery } from "obsidian";
+import { TemplaterError } from "utils/Error";
+
+export class MultiSelectModal<T> extends Modal {
+    private resolve: (values: T[]) => void;
+    private reject: (reason?: TemplaterError) => void;
+    private selectedItems = new Set<string>();
+    private mappedItems = new Map<string, T>();
+    private submitted = false;
+
+    private SELECTOR_SELECTED = ".is-selected";
+    private CLASS_SELECTED = "is-selected";
+
+    private suggestionsContainer: HTMLDivElement;
+    private chipsContainer: HTMLDivElement;
+    private textInput: TextComponent;
+
+    constructor(
+        private text_items: string[] | ((item: T) => string),
+        private items: T[],
+        private unrestricted: boolean,
+        placeholder: string,
+        title: string,
+        private limit?: number
+    ) {
+        super(app);
+
+        // If `T` is of type `string`, then `unrestricted` is set to provided value
+        // else `unrestricted` is always false
+        if (items.length === 0 || typeof items[0] === "string") {
+            this.unrestricted = unrestricted;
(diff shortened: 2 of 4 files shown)


src/core/functions/internal_functions/system/MultiSelectModal.ts
```diff
@@ -0,0 +1,229 @@
...
+        // Map each item to its text representation
+        items.forEach((item) =>
+            this.mappedItems.set(this.getItemText(item), item)
+        );
+
+        // Create Elements
+        this.titleEl.setText(title);
+        this.chipsContainer = this.contentEl.createDiv();
+        this.textInput = new TextComponent(this.contentEl);
+        this.textInput.setPlaceholder(placeholder);
```
AB1908, line 43: Not entirely sure this needs a placeholder. What use case do you have in mind for it?

src/core/functions/internal_functions/system/MultiSelectModal.ts
```diff
@@ -0,0 +1,229 @@
...
+    private submitted = false;
+
+    private SELECTOR_SELECTED = ".is-selected";
+    private CLASS_SELECTED = "is-selected";
+
+    private suggestionsContainer: HTMLDivElement;
+    private chipsContainer: HTMLDivElement;
+    private textInput: TextComponent;
+
+    constructor(
```
AB1908, line 18: Ideally, the constructor should be super simple and the content creation should be delegated to a separate function. The reasoning behind this is that separate functions are easier to test and debug and a constructor should only really take care of initializing of the core components, in this case perhaps just call `super()` like the other modals in the project. Would also be nice for consistency.

AB1908 (commented):
Sorry, didn't have much time to go through the whole thing yet but I do have a few more ideas I'll get to later.

src/core/functions/internal_functions/system/MultiSelectModal.ts
sujyotraut: Sure, I will put all initialization code from constructor to a separate function

src/core/functions/internal_functions/system/MultiSelectModal.ts
sujyotraut: Since `tp.system.suggester` have placeholder so I thought it would be nice to have an option in `tp.system.multiselect` also

sujyotraut:
It add new lines at the start of the file after the result is returned. I can't quite figure out why it's happening or is it just happening in my system, please let me know if it's just happening in my system or same happens on others as well.

https://github.com/SilentVoid13/Templater/assets/70472979/c303a393-37c7-469a-a744-1165bb5b98c8

src/core/functions/internal_functions/system/MultiSelectModal.ts
AB1908: Yes, I think it should be removed

sujyotraut: Should I remove option to set custom title as well?

src/core/functions/internal_functions/system/MultiSelectModal.ts
```diff
@@ -0,0 +1,229 @@
...
+
+                    case "ArrowDown":
+                        this.navigateSelection("DOWN");
+                        break;
+
+                    case "ArrowUp":
+                        this.navigateSelection("UP");
+                        break;
+
+                    case "Enter":
```
Zachatoo, line 74: I'm not sure why, but hitting the <kbd>Enter</kbd> key while in the input is also adding a newline in the editor, causing a conflict and the following notice to pop up.

image

I also feels a bit off to have to click on the input then hit the <kbd>Enter</kbd> key if you're clicking or tapping on options instead of using the keyboard and hitting the <kbd>Tab</kbd> key to select options.

Maybe hitting the <kbd>Enter</kbd> key while the modal is open should close the modal, without having to have the input focused?

Might also be worth adding an "Okay" or "Done" button to make it easier for mobile and touch users.

src/core/functions/internal_functions/system/MultiSelectModal.ts
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    addItem(value: string) {
+        this.selectedItems.add(value);
+        this.addChip(value);
+        this.textInput.setValue("");
+        this.renderSuggestions();
+    }
+
+    removeLastItem() {
```
Zachatoo, line 102: When this function is called, it isn't adding back the item to the list of available suggestions. I think it should be added back as an available item to select.

docs/documentation.toml
```diff
@@ -287,6 +287,35 @@ description = "Throws an error if the prompt is canceled, instead of returning a
 name = "multiline"
 description = "If set to true, the input field will be a multiline textarea"
 
+[tp.system.functions.multiselect]
```
Zachatoo, line 290: This is a tad bit nitpicky, but it may make sense to have this function be called `multi_suggester` to match the naming of the `suggester` function. Which is unfortunate, as I think I'd like `select` and `multiselect` better, but I don't think I want to backtrack on the `suggester`.

Zachatoo (changes requested):
I really appreciate your work on this, I know a few people in the Obsidian Discord have been excited about this. Apologies for not getting to this sooner.

src/core/functions/internal_functions/system/MultiSelectModal.ts
sujyotraut: Do you mean `removeLastItem()` function? 
It removes the selected item and re-renders suggestions by calling `renderSuggestions()`. 
`renderSuggestions()` function filters out items which are selected and renders suggestions which are not selected. 
Since we remove the selected item in `removeLastItem()` it should be available for suggestions.

If you mean something else, can you please provide me more details.

docs/documentation.toml
sujyotraut: Sure, it's no problem I will rename it to `multi_suggester` to make more consistent.

src/core/functions/internal_functions/system/MultiSelectModal.ts
sujyotraut: > I also feels a bit off to have to click on the input then hit the Enter key if you're clicking or tapping on options instead of using the keyboard and hitting the Tab key to select options.

I will work on it so that we don't have to click on input field to select options and add "Done" button to make more mobile friendly.

I am trying to figure out how to fix the issue (adds a newline in the editor) but didn't get any results if you have any suggestions please fill free to let me know.

Zachatoo: I'll try to find some time to figure out why it's adding newlines. I don't think the suggester does this, or if it does, I haven't run into that issue yet.

nirodaonline:
I'm adding a +1 here. 
I'd really like to see this feature.
Thank you both for your work.

PR #1197, opened 2023-09-16
Remove unneeded delay on file creation on newer versions

Zachatoo:
Remove unneeded delay on file creation after version 1.0 of Obsidian.

> I believe that templater can remove that 300ms delay now. The reason was actually a problem in Obsidian core that was fixed a long time ago.
> Basically, the daily notes core plugin would create an empty file, then modify the empty file with your daily note template.
> The issue was that the ‘created’ event was fired before the daily note template was applied, so Templater needed to work around that issue.
> I think that hasn’t been a problem since around 1.0 of Obsidian. Now all the core plugins create the file with the contents already applied in a single filesystem call.

Liam C., Obsidian Team
https://forum.obsidian.md/t/why-i-need-to-wait-a-random-amount-of-time-after-creating-a-file-with-teplater/66066/8

### DIFF
--- a/src/core/Templater.ts
+++ b/src/core/Templater.ts
@@ -2,6 +2,7 @@ import {
     MarkdownPostProcessorContext,
     MarkdownView,
     normalizePath,
+    requireApiVersion,
     TAbstractFile,
     TFile,
     TFolder,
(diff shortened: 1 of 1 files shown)


tkern:
@Zachatoo i just installed the latest release and I think it is the reason it breaks my workflow with Auto Note Mover. Templater fails to template because of:

```
plugin:templater-obsidian:61 Templater Error: Template parsing error, aborting. 
 ENOENT: no such file or directory, open '~/Documents/Notes/Inbox/2023-10-03.md'
```

I create an empty file in Inbox (correct naming though) and then have auto note mover move it to my daily directory. templater is set up with template folders for both `Inbox` and `daily`. without ANM enabled, templater creates the correct inbox note. it also creates the somewhat correct `daily` note IFF i directly create the note in the `daily` folder 

i can see how a delay helped make this work. You reckon it could be related/the reason?

Zachatoo:
Resolved above commend in a Discord conversation. For now, not re-adding the delay, though we may want to consider adding a delay as a configuration setting as a last resort.

https://discord.com/channels/686053708261228577/875720842443649045/1158514236062974022

PR #1215, opened 2023-10-05
Enhance Quote Formatting and Semantic Correctness in `generate_daily_quote`

ShyamGadde:
This PR proposes two changes to the `generate_daily_quote` function:

1. Removal of the `<cite>` element from quote formatting. According to the HTML specification, the <cite> tag defines the title of a creative work (e.g., a book, a poem, a song, a movie, etc.), and it's not intended for citing individuals. Therefore, this change aligns with best practices for HTML and semantic web.

2. Update of quote formatting to use Obsidian callouts. This change provides a more visually appealing and distinguishable format for quotes compared to the generic markdown quote syntax within Obsidian. (With this change, the author’s name stays on a new line, even with strict line breaks enabled.)

Original:

Proposed:

Here are the original and proposed changes:

Original:
```typescript
const new_content = `> ${quote}\n> — <cite>${author}</cite>`;
```

Proposed:
```typescript
const new_content = `> [!quote] ${quote}\n> — ${author}`;
```

### DIFF
--- a/src/core/functions/internal_functions/web/InternalModuleWeb.ts
+++ b/src/core/functions/internal_functions/web/InternalModuleWeb.ts
@@ -37,7 +37,7 @@ export class InternalModuleWeb extends InternalModule {
 
                 const author = json.author;
                 const quote = json.content;
-                const new_content = `> ${quote}\n> — ${author}`;
+                const new_content = `> [!quote] ${quote}\n> — ${author}`;
 
                 return new_content;
             } catch (error) {
--- a/src/core/functions/internal_functions/web/InternalModuleWeb.ts
+++ b/src/core/functions/internal_functions/web/InternalModuleWeb.ts
@@ -37,7 +37,7 @@ export class InternalModuleWeb extends InternalModule {
 
                 const author = json.author;
                 const quote = json.content;
-                const new_content = `> ${quote}\n> — <cite>${author}</cite>`;
+                const new_content = `> ${quote}\n> — ${author}`;
 
                 return new_content;
             } catch (error) {


Zachatoo (approved):

Zachatoo:
I like this. I have tested this and it LGTM.

@AB1908 you think any users would be upset with a change like this? Can always ship it and add a config later if users want to choose how their quotes are formatted.

Marco-Mindstone:
Can we have the choice to use the old version?

Zachatoo:
@Marco-Mindstone I'm hesitant to add in a toggle to have the old version, don't want to end up with a wall of toggles.

Here is a user script that you should be able to drop in replace to get the old style.

daily_quote.js
```js
async function getRequest(url) {
  try {
    const response = await fetch(url);
    if (!response.ok) {
      throw new Error("Error performing GET request");
    }
    return response;
  } catch (error) {
    throw new Error("Error performing GET request");
  }
}

async function daily_quote() {
  try {
    const response = await getRequest("https://api.quotable.io/random");
    const json = await response.json();

    const author = json.author;
    const quote = json.content;
    const new_content = `> ${quote}\n> — <cite>${author}</cite>`;

    return new_content;
  } catch (error) {
    new Error("Error generating daily quote");
    return "Error generating daily quote";
  }
}

module.exports = daily_quote;
```

Usage in template
```
<% tp.user.daily_quote() %>
```

Marco-Mindstone:
@Zachatoo Thank you for your answer.
To be clear, I like the integration itself, the problem is that if you use a Callout "Outline" style this is the result:

image

I wanted to share this with you so if you decide to adjust how it looks on an Outline callout, you can do that.

Thank you for your time!

Zachatoo:
@Marco-Mindstone That doesn't look like a default callout style? I assume that's from a theme or CSS snippet. I don't think Templater should be responsible for making daily quote callouts look good in every theme.

Feel free to modify the code I provided to you to work for callouts instead and look how you'd like 😊

MarkOSullivan94:
@Zachatoo why was this change added? The previous one was more flexible as we could have added `[!quote]` along with our own customisation in a template to make the quotes display nice (which is what I did)

This change should have been in a new function `daily_quote_pretty` rather than breaking existing behaviour for users


### How it looks now

Screenshot 2024-01-08 at 10 07 26



### How it looked previously

Screenshot 2024-01-08 at 10 06 38

Zachatoo:
@MarkOSullivan94 Reasons are in the pull request description.

You can always revert it to the previous behavior with the user script I've provided [here](https://github.com/SilentVoid13/Templater/pull/1215#issuecomment-1815765115) in this same thread.

Zih-Ying:
> @Marco-Mindstone I'm hesitant to add in a toggle to have the old version, don't want to end up with a wall of toggles.
> 
> Here is a user script that you should be able to drop in replace to get the old style.
> 
> daily_quote.js
> 
> ```js
> async function getRequest(url) {
>   try {
>     const response = await fetch(url);
>     if (!response.ok) {
>       throw new Error("Error performing GET request");
>     }
>     return response;
>   } catch (error) {
>     throw new Error("Error performing GET request");
>   }
> }
> 
> async function daily_quote() {
>   try {
>     const response = await getRequest("https://api.quotable.io/random");
>     const json = await response.json();
> 
>     const author = json.author;
>     const quote = json.content;
>     const new_content = `> ${quote}\n> — <cite>${author}</cite>`;
> 
>     return new_content;
>   } catch (error) {
>     new Error("Error generating daily quote");
>     return "Error generating daily quote";
>   }
> }
> 
> module.exports = daily_quote;
> ```
> 
> Usage in template
> 
> ```
> <% tp.user.daily_quote() %>
> ```

I used this, but got the error message: "Error generating daily quote."
Any idea how to fix this? :O

Zachatoo:
@Zih-Ying The quotable API goes down pretty often for long periods of time. Templater has switched to hosting quotes on GitHub [here](https://github.com/Zachatoo/quotes-database). You can do something like this to follow the new way Templater retrieves quotes.

```js
/**
 * Generate a random quote
 * @param {string} url The URL to fetch the quote from
 * @returns A string containing a random quote
 */

async function getRequest(url) {
  try {
    const response = await requestUrl(url);
    if (response.status < 200 && response.status >= 300) {
      throw new Error("Error performing GET request");
    }
    return response;
  } catch (error) {
    throw new Error("Error performing GET request");
  }
}

async function daily_quote() {
  try {
    const response = await getRequest(
      "https://raw.githubusercontent.com/Zachatoo/quotes-database/refs/heads/main/quotes.json"
    );
    const quotes = response.json;
    const random_quote = quotes[Math.floor(Math.random() * quotes.length)];

    const { quote, author } = random_quote;
    const new_content = `> ${quote}\n> — <cite>${author}</cite>`;

    return new_content;
  } catch (error) {
    new Error("Error generating daily quote");
    return "Error generating daily quote";
  }
}

module.exports = daily_quote;
```

PR #1272, opened 2023-12-22
doc: update command utilities prose

nielsbom:
Removed incorrect number, clarified a bit.


Zachatoo (approved):

Zachatoo:
Thanks for making the docs more clear!

Issue #1294, opened 2024-01-12
How does throw_on_cancel work in tp.system.prompt()


Deeds87:
**Plugin information:**

-   OS: [Windows 11]
-   Templater version: [2.1.0]
-   Obsidian version: [1.5.3]

**Describe the bug**
When i use
`let topic = await tp.system.prompt("Set topic", "this is a default value", true, false);`
there is no error in the console and no error modal shown up when i hit "esc".
The file will be created as Untitled.md.

**Expected behavior**
I am expecting the same behaviour as from the suggester comand.
There i get a modal with the cancel error and the file will not be created.

Perhaps i missunterstood how to use throw_on_cancel and there is no bug. If so please help me to unterstand how to achive the desired behaviour.

Thanks for this great plugin

Zachatoo:
I can't reproduce. Here's the script I'm testing with.

```js
<%*
let topic = await tp.system.prompt("Set topic", "this is a default value", true, false);
console.log('I will not be hit due to error');
%>
```

If I execute this script and hit escape or click the X button in the modal, the `console.log` will not be hit because an exception will be thrown.

If I hit enter, then the `console.log` will be hit.

Video below:

https://github.com/SilentVoid13/Templater/assets/6936914/99c424d3-3967-45ad-8ce9-86e6b661e9f4

Deeds87:
I couldn't find my mistake but you are totally right, after trying it again it works like you say.
But what would be more important to me is whether there is a way to interrupt the file creation? Like with suggester.

Zachatoo:
That depends on how the file is being created. There's no way to interrupt the file being created if you're using the `Templater: Create new note from template` command, or if you created the file some other way and Templater is executed because of the `Trigger Templater on new file creation` setting. In both of these cases, the file is created _before_ your template starts executing, it's already too late. The best you could do is delete the file after the fact.

```js
<%*
try {
  const topic = await tp.system.prompt("Set topic", "this is a default value", true, false);
  // other stuff
} catch {
  // if prompt is cancelled (or other error happens), delete file that was created
  const file = tp.file.find_tfile(tp.file.path(true));
  await app.vault.delete(file);
}
-%>
```

You may want to instead consider a two template approach, using the `Templater: Open Insert Template Model` command. This command will not create a new file, it will insert the template into the current file and execute it immediately.

The first template will do any prompts that you need in order to create the file, then create the file.

```js
<%*
[...]

Issue #256, opened 2021-06-02
Cancelled prompt dialog locked the app in mobile


dpurnomo:
**Plugin informations (please complete the following information):**
- OS: [e.g. Windows 10] iOS and iPadOS14
- Templater version: [e.g. 0.3.4] 1.7.1
- Obsidian version: [e.g. 0.9.11] Mobile 0.1.0
- Templater settings: [e.g. Template folder location: "Templates/", Timeout: "5" (add the templates not working with their associated command)]

**Describe the bug**
After triggering a template that ask for input, and cancelling with pressing "Esc" or by tapping outside the box, the dialog did not properly disappear and it locks the whole app. 

**Screenshots**
![66F28A43-D055-4A68-B66B-F54A1EBDF6E9](https://user-images.githubusercontent.com/19817105/120406475-1ffa0100-c375-11eb-9cd7-6e72ba358a8f.png)

insecurejezza:
+1

AB1908:
Could not repro on Android. Can you share the template you used?

PR #1301, opened 2024-01-20
fix: uncaught exception when cancelling prompt when creating note fro…

Zachatoo:
fix: uncaught exception when cancelling prompt when creating note from template and `throw_on_cancel` is true

refs: #1294


Zachatoo:
Reverts #256

PR #1308, opened 2024-01-24
Regenerate pnpm-lock.yaml file to resolve downstream vulnerabilities

andrewmcgivery:
Regenerated pnpm-lock.yaml file in order to upgrade downstream packages to resolve vulnerabilities with those downstream dependencies.

Before:

templaterbefore

After:

templaterafter


Zachatoo:
Thanks for cleaning this up!

PR #1369, opened 2024-05-06
Update Templater.ts

### LINKED ISSUE #1309: Unexpected behavior when using hooks.on_all_templates_executed on a blank file
-   OS: Windows 11
-   Templater version: 2.1.2
-   Obsidian version: 1.5.3
-   Templater settings: default settings with templates folder set to /templates/

**Describe the bug**
When using the hooks.on_all_templates_executed callback like in the [documentation](https://silentvoid13.github.io/Templater/internal-functions/internal-modules/hooks-module.html#hooks-module) with some text that should get added to the note before the frontmatter is updated:
```
<%*
tp.hooks.on_all_templates_executed(async () => {
  const file = tp.file.find_tfile(tp.file.path(true));
  await app.fileManager.processFrontMatter(file, (frontmatter) => {
    frontmatter["key"] = "value";
  });
});
%>

TEXT THAT SHOULD STAY
```
if the note the template is added to is blank (no text, no frontmatter), the text gets added and then removed when the frontmatter is added. By just adding a single character to the note before executing the template, both the text and frontmatter gets added without problem. (There is a pop-up "error" that the file has been modified externally and that it's merging the changes, not sure if that's an issue but you asked about it on discord)

**Expected behavior**
The text and frontmatter should get added to the note even if the note is completely empty.

**Additional context**
I've tried running the template above on two windows 11 computers with the same result, and also tried it in a new empty vault with only templater installed. I'm not 100% sure, but I believe that it worked as expected before, since I have a few templates that rely on this method. I don't use them regularly though, and I updated them to use this method last time I needed to use them (when the hook had just gotten added), so it's possible that I'm misremembering or that the notes I tested them on weren't completely empty.

[link to message on discord](https://discord.com/channels/686053708261228577/1175184210609328229/1200053182244585472)

johnfromoptus:
included check for null active editor before appending blank string. See docs: https://docs.obsidian.md/Reference/TypeScript+API/Workspace/activeEditor

This mitigates edge cases where blank string is appended during transition between active editors. Fixes #1309

### DIFF
--- a/src/core/Templater.ts
+++ b/src/core/Templater.ts
@@ -253,8 +253,9 @@ export class Templater {
         const oldSelections = doc.listSelections();
         doc.replaceSelection(output_content);
         // Refresh editor to ensure properties widget shows after inserting template in blank file
-        await app.vault.append(active_editor.file, "");
-
+        if(active_editor.file){
+          await app.vault.append(active_editor.file, "");
+        }
         app.workspace.trigger("templater:template-appended", {
             view: active_view,
             editor: active_editor,


Zachatoo (approved):
Tests passed and looks good on my end, thanks for this fix!

PR #1379, opened 2024-05-20
docs: Fix broken example of a Script User Function

claremacrae:
The current example fails, giving:
my_functiontp.user.my_script is not a function


claremacrae:
Closed - turns out the name to use is the filename, not the function name.

PR #1385, opened 2024-05-26
feat: new web function to make http requests

Rooyca:
## Short Description

`tp.web` function to make HTTP requests. Optionally, you can specify a path to extract specific data from the response.

### Example

```
<% tp.web.request("https://jsonplaceholder.typicode.com/todos", "0.title") %>
```

Here we are making a call to the `todos` endpoint and we are getting the `title` of the first element.

### DIFF
--- a/src/core/functions/internal_functions/web/InternalModuleWeb.ts
+++ b/src/core/functions/internal_functions/web/InternalModuleWeb.ts
@@ -7,7 +7,7 @@ export class InternalModuleWeb extends InternalModule {
 
     async create_static_templates(): Promise<void> {
         this.static_functions.set("daily_quote", this.generate_daily_quote());
-        this.static_functions.set("request", this.make_request());
+        this.static_functions.set("request", this.generate_request());
         this.static_functions.set(
             "random_picture",
             this.generate_random_picture()
@@ -83,7 +83,7 @@ export class InternalModuleWeb extends InternalModule {
         };
     }
 
-    make_request(): (
+    generate_request(): (
         url: string,
         path?: string,
     ) => Promise<string> {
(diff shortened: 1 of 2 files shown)


src/core/functions/internal_functions/web/InternalModuleWeb.ts
```diff
@@ -80,4 +82,31 @@ export class InternalModuleWeb extends InternalModule {
             }
         };
     }
+
+    make_request(): (
```
Zachatoo, line 86: Can you rename this function to be `generate_request` to match the other function names in this file?

src/core/functions/internal_functions/web/InternalModuleWeb.ts
```diff
@@ -80,4 +82,31 @@ export class InternalModuleWeb extends InternalModule {
...
+                    return path.split('.').reduce((obj, key) => {
+                        if (obj && obj.hasOwnProperty(key)) {
+                            return obj[key];
+                        } else {
+                            throw new Error(`Path ${path} not found in the JSON response`);
+                        }
+                    }, jsonData);
+                }
+
+                return JSON.stringify(jsonData, null, 2);
```
Zachatoo, line 105: What are your thoughts on returning the raw object instead of stringifying it? That way a user could make a single request to get multiple pieces of data. For example, instead of doing this:

```js
Title: <% tp.web.request("https://jsonplaceholder.typicode.com/todos", "0.title") %>
Completed: <% tp.web.request("https://jsonplaceholder.typicode.com/todos", "0.completed") %>
```
A user could do this:
```js
<%* const { title, completed } = await tp.web.request("https://jsonplaceholder.typicode.com/todos", "0"); -%>
Title: <% title %>
Completed: <% completed %>
```

Zachatoo (changes requested):

src/core/functions/internal_functions/web/InternalModuleWeb.ts
Rooyca: Yep, I think that's much better.

Zachatoo (approved):

Zachatoo:
Thank you for your contribution!

PR #1398, opened 2024-06-17
Added copy_from_file and copy_from_folder functions

craigloewen-msft:
This PR adds two functions: `copy_from_file` and `copy_from_folder`.

# Goal 

These functions allow users to take already existing files with full content and copy them to a new location. Some use case for this include: Maintain a record of your notes over time, Copy the latest to do list you have to a new folder so you can update its existing content easily, Copy latest status sync info, 'Template' with existing notes, Use the content you already have in an existing note and keep that up to date, copy new associated files and link to them in your templated note, and more!

# Sample usage

Copy the latest modified file from one folder to another

`<%* await tp.file.copy_from_folder("Status Sync", "lastModified","Status Sync") %>`

Snapshot a particular file and link to it

`Todo snapshot: [[<% (await tp.file.copy_from_file("Todos.md", "Snapshots", "Todo-" + tp.date.now("YYYY-MM-DD"))).basename %>]]`

# Improvements

This PR needs improvements for these areas. If the initial idea looks good please let me know and then I can finish coding in rest of the required items which I believe are:

- [ ] Add documentation
- [ ] Add test coverage

### DIFF
--- a/src/core/Templater.ts
+++ b/src/core/Templater.ts
@@ -221,6 +221,195 @@ export class Templater {
         return created_note;
     }
 
+    async copy_note_from_file(
+        copyfile: string,
+        destfolder?: TFolder | string,
+        filename?: string,
+        open_new_note = true,
+    ): Promise<TFile | undefined> {
+        // TODO: Maybe there is an obsidian API function for that
+        if (!destfolder) {
+            const new_file_location = app.vault.getConfig("newFileLocation");
+            switch (new_file_location) {
+                case "current": {
+                    const active_file = get_active_file(app);
+                    if (active_file) {
+                        destfolder = active_file.parent;
+                    }
+                    break;
+                }
+                case "folder":
+                    destfolder = app.fileManager.getNewFileParent("");
+                    break;
+                case "root":
+                    destfolder = app.vault.getRoot();
+                    break;
+                default:
+                    break;
+            }
+        }
+
+        // Copy file from copyfile to new file in folder with filename if exists
+        const created_note = await errorWrapper(async () => {
+            const copyfile_tfile = resolve_tfile(copyfile);
+            const folderPath = destfolder instanceof TFolder ? destfolder.path : destfolder;
+            const path = app.vault.getAvailablePath(
+                normalizePath(`${folderPath ?? ""}/${filename || copyfile_tfile.basename}`), copyfile_tfile.extension
+            );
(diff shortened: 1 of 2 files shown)


craigloewen-msft:
Can folks help review this? Pinging @Zachatoo ! 

If this functionality doesn't make sense to add in as general functionality I'm also happy to close the PR. 

Thanks!

Zachatoo:
Thank you for your interest in improving Templater!

---

Are there any significant differences between using `tp.file.copy_from_file` and a combination of `tp.file.create_new` and `tp.file.include`? Whether they are or not, I think it should be noted in the documentation. I am open to adding this as a new function.

---

In my opinion, `copy_from_folder` does not pass this criteria for submitting new internal functions. I haven't heard of other users having the use case you have of copying the latest file from one folder to another. If you open an issue and the issue gets a reasonable amount of user feedback that they want it built in, then I think we can proceed with adding this function.

> Keep in mind that only pertinent submissions will be accepted, don't submit a very specific internal variable / function that you'll be the only one using.

craigloewen-msft:
Great question I actually took another stab at this and was able to get that `copy_from_folder` scenario working with Javascript. So I think we can close this if needed as my use case is satisfied, or you can add the functions as you like!

Thank you! :)

Here's the script I used incase it helps others:
```
### Notes
<%*
let fileStringMatch = 'Advanced Paste Weekly Sync';

const files = app.vault.getMarkdownFiles();
let latestFile = null;
let latestTime = 0;
console.log("Starting up", files);
for (const file of files) {
    if (file.basename.includes(fileStringMatch)) {
        console.log("Looking for this file");
        const stats = await app.vault.adapter.stat(file.path);
        if (stats.mtime > latestTime) {
            latestTime = stats.mtime;
            latestFile = file;
        }
    }
}

if (latestFile) {
    let content = await app.vault.read(latestFile);
    const lines = content.split('\n');
    const startIndex = lines.findIndex(line => line.trim() === '### Notes');
    if (startIndex !== -1) {
        const notesContent = lines.slice(startIndex + 1).join('\n');
        tR += notesContent;
    } else {
        tR += "No '### Notes' section found in the file.";
    }
} else {
    tR += "No file found with 'AdvancedPasteWeeklySync' in the title.";
}
%>
<% await tp.file.move("/General Meetings/" + tp.date.now("YYYY-MM-DD") + " Advanced Paste Weekly Sync") %>
```

PR #1418, opened 2024-07-23
Return Relative Paths of Templates in Fuzzy Suggester Prompt

itsonlyjames:
This pull request builds upon the work done by @DraggonFantasy in https://github.com/SilentVoid13/Templater/pull/1078. It includes the following additional changes:
- Removes settings, and removes requirements for settings in `getItemText()` func

Credit to @DraggonFantasy for the initial work.

Fixes https://github.com/SilentVoid13/Templater/issues/1223

PR #1420, opened 2024-07-24
Sort Files by Path

itsonlyjames:
An oversight within https://github.com/SilentVoid13/Templater/pull/1418.

Now sorts suggestor files by path as apposed to solely basename. 

Before:
image

After: 
image

---

This is a great feature, because ultimately, it simply uses the folder structure that matches the templates folder. So if you want to keep a single flat heierarchy, have no folders!


Zachatoo (approved):
Thank you!

PR #1447, opened 2024-09-05
fix(1446): updates docs to match _actual_ behaviour of tp.file.folder()

### LINKED ISSUE #1446: generate_folder() has incorrect `relative` logic vs generate_path()

-   OS: OSX
-   Templater version: head
-   Obsidian version: n/a
-   Templater settings: n/a

**Describe the bug**
The docs and code for `tp.file.folder(relative=false)` are misleading/incorrect.

`folder()` returns a file-relative folder for `relative=false` and a vault-absolute path for `relative=true`. 

This caused some weirdness in my Templated notes where:
- Notes would _sometimes_ (!) be created in the wrong location. 
- When notes were created in the wrong location no template (or the wrong template) would be applied.

```js
// in file `Recurring/Daily/Journal/2024-09-05`
// A link to the following day's Daily-Journal or Daily-Note like:
[[<% tp.file.folder() %>/<% tp.date.now("YYYY-MM-DD", 1, tp.file.title, "YYYY-MM-DD") %>| ⏭️ ]]
// would resolve to:
[[Journal/2024-09-05| ⏭️ ]] // <-- sometime creating the Journal dir in the root of my vault!
// instead of:
[[Recurring/Daily/Journal/2024-09-05| ⏭️ ]] // <-- sometimes creating the Journal dir in the root of my vault!
```

The above snippet is imported/used/shared by multiple templates, so the `folder()` call is quite important./

The issue is quite obvious when you compare `folder()` and `file()` code:

```js
// src/core/functions/internal_functions/file/InternalModuleFile.ts

    // tp.file.folder() - incorrect intended behaviour
    generate_folder(): (relative?: boolean) => string {
            // ... snip ...
            if (relative) {
                folder = parent.path;  // abs for relative 👎
            } else {
                folder = parent.name; // relative for abs 👎
            }
            // ... snip ...
    }

    // tp.file.path() - correct intended behaviour
    generate_path(): (relative: boolean) => string {
            // ... snip ...
            if (relative) {
                return this.config.target_file.path; // relative for relative👍
            } else {
                return `${vault_path}/${this.config.target_file.path}`; // abs for abs 👍
            }
            // ... snip ...
    }

```

**Expected behaviour**
`relative=true` to return Note-relative folder paths, and `relative=false` to return vault-absolute paths.

**Screenshots**
n/a

**Additional context**
Changing or "fixing" the behaviour of the `folder()` API would break back-compat and be a maintenance nightmare, but renaming the value and updating the docs would only affect the docs and improve clarity. However, whilst that is a best-fit for effort/reward, the API itself would be more inconsistent. One possible, and more professional option might be to create a new function `dirname()` (which would match `pathlib` and be less 'surprising' overall, or something better?) and deprecate `folder()`, but is it worth the effort?

doublethefish:
This should clarify behaviour whilst preserving back-compatibility.

This is a best-of solution that will hold until a better one is developed.

fixes #1446

### DIFF
--- a/src/core/functions/internal_functions/file/InternalModuleFile.ts
+++ b/src/core/functions/internal_functions/file/InternalModuleFile.ts
@@ -141,12 +141,12 @@ export class InternalModuleFile extends InternalModule {
         };
     }
 
-    generate_folder(): (relative?: boolean) => string {
-        return (relative = false) => {
+    generate_folder(): (absolute?: boolean) => string {
+        return (absolute = false) => {
             const parent = this.config.target_file.parent;
             let folder;
 
-            if (relative) {
+            if (absolute) {
                 folder = parent.path;
             } else {
                 folder = parent.name;
--- a/docs/documentation.toml
+++ b/docs/documentation.toml
@@ -318,18 +318,18 @@ example = """<% tp.file.find_tfile("MyFile").basename %>"""
 [tp.file.functions.folder]
 name = "folder"
 description = "Retrieves the file's folder name."
-definition = "tp.file.folder(relative: boolean = false)"
+definition = "tp.file.folder(absolute: boolean = false)"
 
 [[tp.file.functions.folder.args]]
-name = "relative"
-description = "If set to `true`, appends the vault relative path to the folder name. If `false`, only retrieves name of folder. Defaults to `false`."
+name = "absolute"
+description = "If set to `true`, returns the vault-absolute path of the folder. If `false`, only returns the basename of the folder (the last part). Defaults to `false`."
 
 [[tp.file.functions.folder.examples]]
 name = "File folder (Folder)"
 example = "<% tp.file.folder() %>"
 
 [[tp.file.functions.folder.examples]]
-name = "File folder with relative path (Path/To/Folder)"
+name = "File folder with vault-absolute path (Path/To/Folder)"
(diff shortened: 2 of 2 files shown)


src/core/functions/internal_functions/file/InternalModuleFile.ts
```diff
@@ -141,12 +141,12 @@ export class InternalModuleFile extends InternalModule {
         };
     }
 
-    generate_folder(): (relative?: boolean) => string {
+    generate_folder(): (absolute?: boolean) => string {
         return (relative = false) => {
```
Zachatoo, line 145: We'll want to update this variable name as well, otherwise I get the following error `Template parsing error, aborting. absolute is not defined`
```suggestion
        return (absolue = false) => {
```

Zachatoo (changes requested):

src/core/functions/internal_functions/file/InternalModuleFile.ts
doublethefish: Oh yes, that was a miss, thank you. I've force-pushed a correctly spelled version.

Zachatoo: Perfect, thanks for fixing that!

Zachatoo (approved):
Thanks for the contribution!

PR #1456, opened 2024-09-13
Allow specifying custom extension in create_file

gergesh:


Zachatoo:
Thanks for contributing!

Can you help me understand the use case? I believe `tp.file.create_new` should work as is when referencing a template file that is of any file type, without having to explicitly define the extension.

src/core/Templater.ts
```diff
@@ -136,8 +137,8 @@ export class Templater {
             }
         }
 
-        const extension =
```
gergesh, line 139: This isn't the case, as you can tell from this line that overrides the extension in the newly created note

gergesh:
To be specific on my usage: I'm simply trying to create a new file with a custom extension from a template script.

src/core/Templater.ts
Zachatoo: Only if you're providing a string template, as opposed to providing a file for the template. If you're providing a file then it will use the same extension as the provided file. I've used this for .canvas files in the past. Are you generating a string in code and creating a file using that generated string?

I'm hesitant to add yet another argument to this function if it isn't strictly needed, it has a lot of arguments already, and I don't suspect this argument would be used much since most use cases can already be covered.

gergesh: My usecase isn't too niche (I believe), I have a script that prompts for a filename and creates a file named "yyyy-mm-dd-filename" (which I'd like to be able to support arbitrary extensions)

PR #1459, opened 2024-09-19
fix: Change API provider for daily quotes

### LINKED ISSUE #1407: Daily Quote errors

-   OS: Fedora 40
-   Templater version: 2.3.3
-   Obsidian version: 1.6.5

**Describe the bug**
when using `<% tp.web.daily_quote() %>` it errors out with the message 

**Expected behavior**
Should generate a quote

**Screenshots**
image

**Additional context**
Started to since yesterday, worked after a few tries, today even after 10 tries no generation of daily quote

Best Regards,
Kalyan Mudumby

chance-nelson:
## Overview
The Quotable API has been unreachable due to SSL cert expiration since approximately June, with no sign of progress from the package matainers (see [Issues in lukePeavey/quotable)](https://github.com/lukePeavey/quotable/issues?q=is%3Aissue+is%3Aopen+label%3A%22%3Awarning%3A+API+is+down%22)). To restore functionality for daily quotes, this patch changes the provider API over to [Zenquotes](https://docs.zenquotes.io/zenquotes-documentation/#api-structure).

Resolves #1407 

## Notes
- This patch uses the `/daily` endpoint from Zenquotes, which does not perfectly match the functionality of the random Quotable endpoint, in the manner that it returns the same quote for a given 24 hour timespan. This seems to match more with the intended use of this function, but if there is concern about keeping quotes perfectly random per call to Templater, this can be switched over to `/random`

### DIFF
--- a/src/core/functions/internal_functions/web/InternalModuleWeb.ts
+++ b/src/core/functions/internal_functions/web/InternalModuleWeb.ts
@@ -34,12 +34,12 @@ export class InternalModuleWeb extends InternalModule {
         return async () => {
             try {
                 const response = await this.getRequest(
-                    "https://api.quotable.io/random"
+                    "https://zenquotes.io/api/today"
                 );
                 const json = await response.json();
 
-                const author = json.author;
-                const quote = json.content;
+                const author = json.a;
+                const quote = json.q;
                 const new_content = `> [!quote] ${quote}\n> — ${author}`;
 
                 return new_content;


Zachatoo:
Thanks for the contribution!

In it's current state, it doesn't appear to work for a few reasons:
1. That API has a CORS policy that prevents Obsidian from fetching the quote data. We can get around this by using [requestUrl](https://docs.obsidian.md/Reference/TypeScript+API/requestUrl) instead of `fetch`. I suspect I need to make this change anyways to fix other issues.
2. The response from the `/daily` endpoint returns an array with a single object. The code will need to be updated to account for that.

On a more legal note, if we use this API we will be required to link back to their website since we're using the free version of this API, as per https://docs.zenquotes.io/zenquotes-documentation/#use-limits. So if we want to use this API we will need to include the text they provided somewhere in the quote. I'm not totally sure we want to use an API that requires attribution, as it will throw a lot of users off.

I'll be continuing the discussion around which API provider to use in the related issue, feel free to share your thoughts on which API provider to use there!

Zachatoo (changes requested):
See other comment for requested changes.

Zachatoo:
Closing in favor of using https://github.com/Zachatoo/quotes-database for the new quotes API

PR #1468, opened 2024-10-01
Streamline folder template UX

### LINKED ISSUE #1428: Suggestion to Improve New Folder Template Addition Behavior in Templater Settings
Is your feature request related to a problem?
No

Describe the solution you'd like:
In the Templater settings, adding a new folder template using the + button currently inserts the new item at the bottom of the list and automatically scrolls the page up. This behavior can be improved in two ways:

Solution 1: The button should add the new item just below it (at the first index instead of the last index) and focus on it without scrolling.

Solution 2: The button should add the new item at the bottom (current behavior) and scroll down to focus on the inserted item.

Describe alternatives you've considered:
No

Additional context:
This request pertains to the Templater/Folder Templates settings.

madigan:
Small quality of life issues related to folder templates:

1. Change the "+" button to be the same as the other "Add" buttons on the page (resolves #1428).
2. There were two headers being created for the "Folder templates" section; remove the duplicate.
3. The folder templates section is only visible if templates can be triggered on file create, but the toggle is further up in the settings. Move the Folder Templates section to be directly underneath the toggle so the user can easily see what changed.

### DIFF
--- a/src/settings/Settings.ts
+++ b/src/settings/Settings.ts
@@ -1,11 +1,10 @@
+import TemplaterPlugin from "main";
 import { ButtonComponent, PluginSettingTab, Setting } from "obsidian";
 import { errorWrapperSync, TemplaterError } from "utils/Error";
-import { FolderSuggest } from "./suggesters/FolderSuggester";
-import { FileSuggest, FileSuggestMode } from "./suggesters/FileSuggester";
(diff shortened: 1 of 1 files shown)


Zachatoo (approved):
Love it! Thanks for contributing to Templater!

PR #1475, opened 2024-10-03
Add insert hotkey setting

### LINKED ISSUE #1361: Create new file from template
### Discussed in https://github.com/SilentVoid13/Templater/discussions/514

<div type='discussions-op-text'>

<sup>Originally posted by **Ampa** April  3, 2021</sup>
At present we can create a new file, and then as a second step insert a template.

Would you consider a new action: Create new from template?

This would carry out both steps as a single action.
</div>

It is 2024 and it still does'nt work.

My settings:

Screenshot 2024-04-26 at 16 13 22

When there is no active note in Obsidian and i will trigger the hotkey im getting this:

Screenshot 2024-04-26 at 16 09 12

And this is the template:

Screenshot 2024-04-26 at 16 15 18

There is no append command in the code!

madigan:
- Fix bug where the hotkey search isn't filtered to the command
- Changed the hotkey search to be filtered by template (not inserts in general)
- Generate create command as well as insert command (Resolves #1361)

### DIFF
--- a/src/handlers/CommandHandler.ts
+++ b/src/handlers/CommandHandler.ts
@@ -1,6 +1,6 @@
 import TemplaterPlugin from "main";
-import { resolve_tfile } from "utils/Utils";
 import { errorWrapperSync } from "utils/Error";
+import { resolve_tfile } from "utils/Utils";
 
 export class CommandHandler {
     constructor(private plugin: TemplaterPlugin) {}
@@ -79,13 +79,14 @@ export class CommandHandler {
 
     add_template_hotkey(
         old_template: string | null,
-        new_template: string
+        new_template: string,
+        mode: string
     ): void {
         this.remove_template_hotkey(old_template);
 
         if (new_template) {
             this.plugin.addCommand({
-                id: new_template,
+                id: `insert-${new_template}`,
                 name: `Insert ${new_template}`,
                 icon: "templater-icon",
                 callback: () => {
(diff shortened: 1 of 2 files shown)


src/settings/Settings.ts
```diff
@@ -271,7 +271,7 @@ export class TemplaterSettingTab extends PluginSettingTab {
                                 app.setting.openTabById("hotkeys");
                                 // @ts-ignore
                                 const tab = app.setting.activeTab;
-                                tab.searchInputEl.value = "Templater: Insert";
+                                tab.searchComponent.inputEl.value = template;
```
Zachatoo, line 274: Nice catch on this!

src/handlers/CommandHandler.ts
```diff
@@ -79,13 +79,14 @@ export class CommandHandler {
 
     add_template_hotkey(
         old_template: string | null,
-        new_template: string
+        new_template: string,
+        mode: string
```
Zachatoo, line 83: Can you remove this since we aren't using it?

src/handlers/CommandHandler.ts
```diff
@@ -79,13 +79,14 @@ export class CommandHandler {
...
     ): void {
         this.remove_template_hotkey(old_template);
 
         if (new_template) {
             this.plugin.addCommand({
-                id: new_template,
+                id: `insert-${new_template}`,
```
Zachatoo, line 89: Note to self: make sure changelog includes note about this breaking change. I don't think many people are using this, and I like that this is less ambiguous.

Zachatoo (changes requested):

src/handlers/CommandHandler.ts
Zachatoo: @madigan This is the only actual change that I'd like to see, this looks great!

Zachatoo: Eh, I can fix it real quick

Issue #1480, opened 2024-10-09
Update to 2.8.0 removed all Template hotkeys


tyleralbee:
**Plugin information:**

-   OS: Ubuntu 22.04.5 LTS
-   Templater version: 2.8.0
-   Obsidian version: 1.6.7
-   Templater settings: 

```json
{
  "command_timeout": 5,
  "templates_folder": "templates",
  "templates_pairs": [
    [
      "",
      ""
    ]
  ],
  "trigger_on_file_creation": true,
  "auto_jump_to_cursor": true,
  "enable_system_commands": false,
  "shell_path": "",
  "user_scripts_folder": "",
  "enable_folder_templates": true,
  "folder_templates": [
    {
      "folder": "orgs/redacted",
      "template": "templates/redacted.md"
    },
    {
      "folder": "orgs/example/redacted-2",
      "template": "templates/daily-example-redacted-2.md"
    }
  ],
  "enable_file_templates": false,
  "file_templates": [
    {
      "regex": ".*",
      "template": ""
    }
  ],
  "syntax_highlighting": true,
  "syntax_highlighting_mobile": false,
  "enabled_templates_hotkeys": [
    "templates/datetime.md",
    "templates/history.md",
    "templates/info.md",
    "templates/invoice/lines/code.md",
    "templates/invoice/lines/meeting.md",
    "templates/time.md",
    "templates/task.md"
  ],
  "startup_templates": [
    ""
  ]
}
```

**Describe the bug**

After updating Templater to 2.8.0, all of my custom Template hotkeys were wiped. To be clear, the plugin's standard hotkeys were not removed; only hotkeys that were connected to my custom templates. This includes all of the hotkeys that were connected to templates defined in `"enabled_templates_hotkeys"`. In my case, this consisted of:

- Templater: Insert templates/datetime.md
- Templater: Insert templates/history.md
- Templater: Insert templates/info.md
- Templater: Insert templates/invoice/lines/code.md
- Templater: Insert templates/invoice/lines/meeting.md
- Templater: Insert templates/time.md
- Templater: Insert templates/task.md

Hotkeys that were **not** removed:

- Templater: Create new note from template
- Templater: Jump to next cursor location
- Templater: Open insert template modal
- Templater: Replace templates in the active file

I would imagine that the following hotkeys **would have been** removed (but I did not have them set):

- Templater: Create templates/datetime.md
- Templater: Create templates/history.md
- Templater: Create templates/info.md
[...]

PR #1482, opened 2024-10-11
Fix insert command removal

### LINKED ISSUE #1481: New hotkeys are being created instead of changing path of the existing one
-   OS: Ubuntu 22.04
-   Templater version: 2.8.1
-   Obsidian version: 1.6.7
-   Templater settings: None

**Describe the bug**
Creates many unnecessary hotkeys

**Expected behavior**
As I change the path of a template for an existing hotkey, I expect it to change the path of the existing one. Instead as I type each character, new hotkey slots are being created.

**Screenshots**
Screenshot from 2024-10-11 11-57-28

madigan:
Fixed bug where "insert" commands were added without a prefix, but removed with a prefix (see #1480; resolves #1481).

### DIFF
--- a/src/handlers/CommandHandler.ts
+++ b/src/handlers/CommandHandler.ts
@@ -129,7 +129,7 @@ export class CommandHandler {
                 `${this.plugin.manifest.id}:create-${template}`
             );
             app.commands.removeCommand(
-                `${this.plugin.manifest.id}:insert-${template}`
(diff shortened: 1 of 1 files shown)


Zachatoo (approved):
Thanks for catching this, this was my bad!

PAAtchison:
@madigan I am still experiencing this issue after re-installing Templater.

PR #1494, opened 2024-10-31
Adding setting option for renaming new files on creation.

ChloBearDesigns:
This will allow users to disable the auto rename on file creation. It's annoying if you design your templates to already do the renaming. This will allow users to disable it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,5 @@ Icon
 Network Trash Folder
 Temporary Items
 .apdisk
+
+package-lock.json
--- a/src/core/Templater.ts
+++ b/src/core/Templater.ts
@@ -218,9 +218,11 @@ export class Templater {
                 true
             );
 
-            active_leaf.setEphemeralState({
-                rename: "all",
-            });
+            if (this.plugin.settings.prompt_rename_on_creation) {
+                active_leaf.setEphemeralState({
+                    rename: "all",
+                });
+            }
         }
 
         await this.end_templater_task(path);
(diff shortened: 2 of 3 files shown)


ChloBearDesigns:
Hey, just adding this quick setting to toggle on the auto file rename when creating new files. I was annoying that I already do this in my templates and it's not needed. I suspect others might like this too, if not.. no worries.

Zachatoo:
I think this is a great idea, thank you for putting this together! I have a few thoughts:

1. Is there any way to have `tp.file.title` populated with the user entered file name? Otherwise you wouldn't be able to use `tp.file.title` in the template and have it be accurate.
2. I'd like this setting to be defaulted to off, to avoid the support headache that will come from users seeing duplicate prompts for users who already have a prompt.
3. We should rename the setting name to "Prompt to rename on create new note from template" and update the description to "Templater will prompt you to rename the file when creating a new note from a template."
4. Can we remove the changes to the package-lock.json file, since it isn't related to this feature?

ChloBearDesigns:
hey Zachatoo, sure I can make those changes. Sorry for not getting back to you sooner, Github didn't tell me I had any messages until I went to look!

ChloBearDesigns:
I made the changes to update the setting name and I defaulted it to off, but I had a quesiton about you first point:
`"Is there any way to have tp.file.title populated with the user entered file name? Otherwise you wouldn't be able to use tp.file.title in the template and have it be accurate."`
I'm not sure what you mean by this. All my change does is bypass the active_leaf.setEphemeralState() call after the file is created. my use case is that I manually renamed the filename from inside the templater script. Either way, tp.file.title wouldn't be correct since the new file name is already created and the rename happened after all templater scripts were finished.

Zachatoo:
Cleaning up stale PRs.

PR #1546, opened 2025-02-27
fix(docs): Format argument as optional parameter

mdrizwanfk:


Zachatoo (approved):

Zachatoo:
Thank you for fixing this!

PR #1539, opened 2025-02-08
Feature/add user script descriptions

srz2:
This is a feature I've wanted in templater as something that would help me.

- Implemented a new feature that adds 'intellisense' like support for user scripts. This allows descriptions, arguments, and return descriptions
- Added `user` to the **tp** model to allow easier access to user scripts. This is conditional if _Templater_ detects user scripts
- Added quick documentation  for this feature

# Example of user 'intellisense'
Screenshot 2025-02-08 at 1 35 54 AM

# Example of user script (preserves normal look when nothing provided)
Screenshot 2025-02-11 at 11 54 08 AM

# Example of user script 'intellisense' with description provided
Screenshot 2025-02-08 at 1 36 50 AM

# Example of user script with return and arguments
Screenshot 2025-02-11 at 11 50 32 AM
 'intellisense'

PR #1551, opened 2025-03-07
Feature/intellisense render settings

srz2:
This adds a feature to allow render settings for intellisense which was added in #1539 
I thought it could be beneficial to alter the render despite what is provided in the user script file

This adds a new setting
 
Screenshot 2025-03-06 at 11 01 48 PM

This allows the user to change the render from the DEFAULT of full to various options including off.

This allows the user to render what they want instead of having to alter the source code of the user script which they might not want to do

## Off
Screenshot 2025-03-06 at 11 09 11 PM

## Full Render
Screenshot 2025-03-06 at 11 09 22 PM

## Description and Parameter List
Screenshot 2025-03-06 at 11 09 33 PM

## Description and Return
Screenshot 2025-03-06 at 11 09 43 PM

## Description Only
Screenshot 2025-03-06 at 11 09 52 PM

### DIFF
--- a/src/editor/Autocomplete.ts
+++ b/src/editor/Autocomplete.ts
@@ -165,4 +165,8 @@ export class Autocomplete extends EditorSuggest<TpSuggestDocumentation> {
             return 0;
         }
     }
+
+    updateAutocompleteIntellisenseSetting(value: IntellisenseRenderOption){
+        this.intellisense_render_setting = value;
+    }
 }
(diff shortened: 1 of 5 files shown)


srz2:
Hi @Zachatoo i had one more suggestion for the intellisense

Zachatoo (approved):
Looks great! Thanks for this addition!

Issue #1518, opened 2024-12-31
Feature Request: Support multiple template folders


JesseTG:
**Is your feature request related to a problem? Please describe.**

I'd like to be able to keep my templates in the areas where they'll actually be used.

**Describe the solution you'd like**

The ability to set multiple template folders, instead of just one.

**Describe alternatives you've considered**

I currently keep all of my templates in one folder; the problem is that as I continue to write more of them, bouncing between the folders for my templates and my actual notes becomes tedious.

**Additional context**

I adopt the [Johnny.Decimal](https://johnnydecimal.com) system for organizing my knowledge, documents, and tasks. Each section has its own "meta" area that's used to manage the section itself; the one relevant to this ticket is an area for templates, as described [here](https://johnnydecimal.com/10-19-concepts/12-advanced/12.03-the-standard-zeros/#05-templates).

rolemartyr-x:
+1 to this idea - I'm implementing the Johnny.Decimal now and am running into this exact issue.

PR #1555, opened 2025-03-14
Add support for multiple template folders

GGonnerman:
Implements #1518

### DIFF
--- a/src/core/Templater.ts
+++ b/src/core/Templater.ts
@@ -462,6 +462,40 @@ export class Templater {
         }
     }
 
+    get_templates_folders(): Array<string> {
+        let template_folders: Array<string>
+        if(this.plugin.settings.enable_multiple_template_folders) {
+            template_folders = this.plugin.settings.templates_folders;
+        } else {
+            template_folders = [this.plugin.settings.templates_folder];
+        }
+
+        return template_folders.filter((folder: string) => folder !== "");
+    }
+
+    get_templates_folders_shared_path(): string {
+        if(this.plugin.templater.get_templates_folders().length === 0) {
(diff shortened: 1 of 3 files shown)


GGonnerman:
Just to confirm my understanding, your recommendation would be to:

1. Have 2 options (templates_folder and templates_folders)
2. Have a 1-time migration that copies the value from templates_folder to templates_folders
3. Clear the templates_folder value
4. Rewrite the code to assume that templates_folders is used everywhere and templates_folder is deprecated

You mentioned leaving the data for templates_folder alone to support legacy users, but I don't see how to do this without issues. There doesn't seem to be a way to set a default value for a setting (that is dependent on another setting's value). Instead, if we manually check if templates_folders is empty and copy the value from templates_folder, the migration would trigger each time the plugin is launched and templates_folders is empty (which would not be expected user behavior).

Does this sound correct? Or do I have a misunderstanding somewhere?

Zachatoo:
> 3. Clear the templates_folder value

Why do we need to clear this? The only thing we need to check is if `templates_folders` is undefined, if so, copy over `templates_folder`. We can do this in the [load_settings()](https://github.com/SilentVoid13/Templater/blob/4ab3bd5016ead499fbb878bb1b28342789250453/src/main.ts#L67-L73) method.

> The setting templates_folders can be undefined, it could be defaulted to a string that is for example "uninitiated_array". Then when the plugin loads check for a non existent setting or templates_folders =="uninitiated_array a setting that isn't an array.

I recommend using `undefined` instead of a string, optionally with a JSDoc comment explaining what it means if it's `undefined`.

PR #1589, opened 2025-06-12
Cleaner display of custom hot key in command palette

sualk01:
The name of a custom command, which inserts or creates a template, shows now the title of the template instead of its path, which improves readability in the command palette.

### DIFF
--- a/src/handlers/CommandHandler.ts
+++ b/src/handlers/CommandHandler.ts
@@ -93,7 +93,7 @@ export class CommandHandler {
         if (new_template) {
             this.plugin.addCommand({
                 id: new_template,
-                name: `Insert ${new_template}`,
+                name: `Insert ${new_template.split(/\/|\./).slice(-2,-1)}`,
                 icon: "templater-icon",
                 callback: () => {
                     const template = errorWrapperSync(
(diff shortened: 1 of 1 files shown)


Zachatoo (changes requested):
I don't mind removing the templates folder from the command, but we should preserve subfolders. For example, a template file at Templates/folder/template.md should show as folder/template

src/handlers/CommandHandler.ts
```diff
@@ -93,7 +94,7 @@ export class CommandHandler {
         if (new_template) {
             this.plugin.addCommand({
                 id: new_template,
-                name: `Insert ${new_template}`,
+                name: `Insert ${new_template.slice(Settings.templates_folder.length,-3)}`,
```
Zachatoo, line 97: The build is failing with this latest change

Zachatoo (changes requested):

src/handlers/CommandHandler.ts
```diff
@@ -92,9 +91,13 @@ export class CommandHandler {
...
 
         if (new_template) {
+            const new_template_name = new_template.slice(
+                this.plugin.settings.templates_folder.length + 1,
+                -3
+            );
             this.plugin.addCommand({
-                id: new_template,
-                name: `Insert ${new_template.slice(Settings.templates_folder.length,-3)}`,
+                id: new_template_name,
```
Zachatoo, line 99: Let's keep the `id` the same so we don't break any keybinds with this update.

Zachatoo (changes requested):

Zachatoo (approved):

PR #1601, opened 2025-07-06
feat: merge frontmatter properties when inserting a template & allow templater to work with Bases core plugin

### LINKED ISSUE #1218: Merge properties in the template with properties in the note
The Templates core plug-in now merges the YAML properties in the template with the ones already in the note. 

Template:
```
---
category: test
tags: 
  - test
---

Some test text
```

Note:
```
---
tags: 
  - draft
---
This is my text
```

When I apply the template, using the core plug-in, the text in the note becomes:
```
---
tags: 
  - draft
  - test
category: test
---
This is my text

Some test text
```

If I do the same with Templater, what I get is:
```
---
tags: 
  - draft
---
This is my text
---
category: test
tags: 
  - test
---

Some test text
```

_Originally posted by @paolo-santucci in https://github.com/SilentVoid13/Templater/issues/1191#issuecomment-1741874483_

### LINKED ISSUE #1387: Templater should parse, merge and write out frontmatter gracefully
**Is your feature request related to a problem?** 
Many of the recent bug reports and feature requests concern managing and updating frontmatter/properties. `app.fileManager.processFrontmatter()` has been proposed as a workaround for some of those problems, but has
- brought with it its own issues (especially regarding race conditions)
- failed to provide solutions for some of them (retrieving, parsing and applying frontmatter from the yaml block of a simple template file, for example)
- left most of the work to the end user

I propose that Templater implements parsing and handling of frontmatter properties, whether they are initially defined as objects (compare `console.log(tp.frontmatter)`)

```js
{
    "modified": "2024-05-29T16:36:31+02:00",
    "tags": ["source"]
}
```

or as YAML strings

```yaml
`---
modified: "2024-05-29T16:36:31+02:00",
tags: ["source"]
---`
```

Usage Example: Modularization of Template files. For example, I might want to add certain properties to all notes on "Beliefs" to define how _strongly_ I believe in something and _why_, and also add a little section to add further comments. Let's call that an `Epistemology Module`.

My note on a belief:
Cows are mammals.md
```md 
---
created: 2020-12-12
tags: [#belief]
---
# Cows are mammals

I believe that cows are mammals. Because.
```

I only learned about epistemology statements in 2024, but have started adding them to all my beliefs:

epistemology-module.md
```md
---
tags: [#belief, #epistemology]
credence: 60
effort: 30
---

## Epistemology statement
How strongly do I believe this? Hint: When creating a file, the default credence is <% tp.frontmatter["credence"] %>
How much effort did I put into this?
When did I last update this statement? <% tp.date.now() %>
```

I now want to `tp.file.include("[[epistemology-module.md]]")` in my `belief-template.md` template for new beliefs and also be able to `Insert template` into existing ones.

When I create a new file using epistemology-module.md, it creates a new file with the correct YAML block just fine, but still fails on inserting the tp.frontmatter["credence"] (because the frontmatter of the new file is not defined yet, even though it is written out in the template). When I add that epistemology-module.md to the existing belief, it inserts the YAML-block at cursor position, not processing it at all, and obviously still fails inserting the credence.

To fix this, all frontmatter that is handled during execution of a script should:
- Be parsed from YAML to an object (possibly replacing `<% tp.stuff %>` where appropriate).
  - This is important during `tp.file.include(tFile)` or when Inserting a template to an existing file when the template in question contains its own metadata block and ...
  - ... useful for filling tp.frontmatter during creation of a new file
- Be merged gracefully when pulling frontmatter from more than one source (due to a non-empty file, several `tp.file.include`, etc.):
  - new keys should be added
  - duplicate keys should be overwritten with the new value if that value is a primitive...
  - ... or merged if the value itself is an object/array
- Be written out as a unified YAML-block at the top

Upon inserting the module into the existing file, we should thus get:

Cows are mammals.md
```md 
---
created: 2020-12-12
tags: [#belief, #epistemology]
credence: 60
effort: 30
---
# Cows are mammals

I believe that cows are mammals. Because.
## Epistemology statement
How strongly do I believe this? Hint: When creating a file, the default credence is 60
How much effort did I put into this?
When did I last update this statement? 2024-05-29
```

Currently, I use a haphazardly hacked-together mergeFrontmatter() function, write weird JS-Object definitions for my templates and modules (instead of simple .md templates) and then create a template literal containing the template to use for `tp.file.create_new()`. My module-definitions have a method() for updating an existing file with that module (e.g.: an excalidraw-module would be adding the `excalidraw-plugin: parsed` property using `processFrontmatter()` and the Excalidraw body `%%\n# Excalidraw Data\n## Text Elements [...]` by inserting that literal at the cursor position). I have not found a way to parse the template literals through Templater in the same step, yet (because only `tp.file.create_new()` accepts a string for a template and Templater does not currently expose its `parse_template()` function). Therefore, if my template or module definition includes eta-tags, I hit my "Replace templates in the active file" hotkey afterwards.

**Additional context**
Both the built-in Templates-Plugin as well as the Excalidraw-Plugin parse templates of the format 

```md
---
excalidraw-plugin: parsed
---

# Excalidraw body ....
```

without any problems and can create a _new_ file from such a template, or _add_ that template to an existing file (merging the new key into an existing file's frontmatter).

### LINKED ISSUE #1600: Insert template from new note created by Base

When you create a new note from a `Base`, the note has some frontmatter properties set depending on the filters used in the base. This means that templater does not trigger the template on new file creation since the file is not empty.

I would like `Trigger Templater on new file creation` to work with files created using the `New Note` feature in the core plugin `Bases`

Ideally the frontmatter properties created by the base would be left in tact, with new frontmatter properties added by the template added to the frontmatter, and property that has a value in the template should overwrite the base property. If the value is blank in the template then the property should be added if it was not created by the base, if it was added by the base then the value should be left. If the property is a list then the values in the template should be appended to the property if it was created by the base.

The above functionality replicates the behaviour of the core `Templates: Insert Template` command wrt how properties are handled.

I tried removing the `file.stat.size == 0` check in `on_file_creation()`
https://github.com/SilentVoid13/Templater/blob/5d8f9f0430f8a31828a77c9bba8640473c794716/src/core/Templater.ts#L515
This allows templater to run when a file is created using a base but it has several issues
- Copy and pasting files is broken as the contents will be overwritten on paste by the template
- The properties created by the base are completely overwritten by the template

**Additional context** Add any other context or screenshots about the feature
request here.
I was working on a way to save the frontmatter created by the base so it could be restored in the template and came up with this script.

```js
<%*
const file = tp.file.find_tfile(tp.file.path(true));
console.log(file)
let fm = null;

// Save frontmatter generated by Obsidian base
await app.fileManager.processFrontMatter(file, (frontmatter) => {
    fm = frontmatter;
});

// Restore Obsidian base frontmatter
tp.hooks.on_all_templates_executed(async () => {
  const file = tp.file.find_tfile(tp.file.path(true));
  await app.fileManager.processFrontMatter(file, (frontmatter) => {
    Object.keys(fm).forEach(key => {
      frontmatter[key] = fm[key];
    });
  });
});

%>
```

It doesn't fix all the problems (ie copy and paste of files is still broken), and it doesn't handle additional properties set in the template.

My thinking is rather than trying to do this in the template itself, it would be better suited to be done by the templater plugin.

What are your thoughts on this issue?

I can't find any reference of people using templates with new notes created with bases since it is such a new feature

AndyEveritt:
2 main feature additions here:
Fixes #1600 
Fixes #1218 
Fixes #1387

# Merging frontmatter properties
It has long been an issue that templater could not merge frontmatter properties when inserting a template like the core `Templates: Insert Template` command. Issues have discussed this feature request #1218, #1387, & #1600.

Now if a template is inserted into a file, the template is processed and any frontmatter found in the resulting template is merged into the file

If a template is included in another template using `tp.file.include()` and both the original file, template, and sub template have frontmatters, they will all be merged into a single frontmatter.

I have tried to keep the merging rules consistent with how `Templates: Insert Template` works.

# Templater now works with Obsidian Bases
Bases can be used to create new notes with some frontmatter prepopulated. If the new note is created in a folder where Templater would usually automatically insert a template, it previously would not work as there was a check in Templater to ensure the file is empty before inserting the template.

Now the check looks to see if the file is empty excluding the frontmatter. So templater will correctly insert the template when a new note is created using a base, and the frontmatter will be merged correctly between the base generated note and the template

### DIFF
--- a/src/utils/Utils.ts
+++ b/src/utils/Utils.ts
@@ -250,92 +250,39 @@ export async function merge_front_matter(
         return;
     }
     try {
-        await Promise.all(
-            Object.keys(properties).map(async (prop) => {
-                const value = properties[prop];
-                if (app.metadataCache.getFileCache(file)?.frontmatter == null) {
-                    // console.log("Frontmatter is empty");
-                    await app.fileManager.processFrontMatter(
-                        file,
-                        (frontmatter) => {
-                            // console.log(
-                            //     `adding new property: ${prop} to ${file.basename} with value: ${value}`
-                            // );
-                            frontmatter[prop] = value;
-                        }
-                    );
+        await app.fileManager.processFrontMatter(file, (frontmatter) => {
+            for (const prop in properties) {
+                const currentValue = frontmatter[prop];
+                const newValue = properties[prop];
+
+                if (currentValue === undefined) {
+                    // If the property doesn't exist, add it
+                    frontmatter[prop] = newValue;
                 } else if (
-                    app.metadataCache
-                        .getFileCache(file)
-                        ?.frontmatter?.hasOwnProperty(prop)
+                    Array.isArray(currentValue) ||
+                    Array.isArray(newValue)
                 ) {
-                    // console.log(`${file.basename} contains property: ${prop}`);
-                    const originalValue =
-                        app.metadataCache.getFileCache(file)?.frontmatter?.[
-                            prop
-                        ];
-                    if (
-                        value != null &&
-                        (Array.isArray(originalValue) || Array.isArray(value))
-                    ) {
-                        // console.log(`${prop} is an array`);
-                        await app.fileManager.processFrontMatter(
-                            file,
-                            (frontmatter) => {
-                                if (!Array.isArray(originalValue)) {
-                                    // console.log(
-                                    //     `converting ${prop} to an array in ${file.basename}`
-                                    // );
-                                    frontmatter[prop] = [originalValue];
-                                }
-                                if (Array.isArray(value)) {
-                                    for (let i = 0; i < value.length; i++) {
-                                        // console.log(
-                                        //     `adding ${value[i]} to ${prop} in ${file.basename}`
-                                        // );
-                                        if (
-                                            !frontmatter[prop].includes(
-                                                value[i]
-                                            )
-                                        ) {
-                                            frontmatter[prop].push(value[i]);
-                                        }
-                                    }
-                                } else {
-                                    // console.log(
-                                    //     `adding ${value} to ${prop} in ${file.basename}`
-                                    // );
-                                    if (!frontmatter[prop].includes(value)) {
-                                        frontmatter[prop].push(value);
-                                    }
-                                }
-                            }
-                        );
-                    } else if (originalValue !== value && value != null) {
-                        // console.log(
-                        //     `updating property: ${prop} in ${file.basename} from ${originalValue} to ${value}`
-                        // );
-                        await app.fileManager.processFrontMatter(
-                            file,
-                            (frontmatter) => {
-                                frontmatter[prop] = value;
-                            }
-                        );
-                    }
-                } else {
-                    // console.log(`${file.basename} doesn't contain ${prop}`);
-                    // console.log(
-                    //     `adding property: ${prop} to ${file.basename} with value: ${value}`
-                    // );
-                    await app.fileManager.processFrontMatter(
-                        file,
-                        (frontmatter) => {
-                            frontmatter[prop] = value;
-                        }
+                    // If either is an array, merge them
+                    frontmatter[prop] = Array.from(
+                        new Set([
+                            ...(currentValue
+                                ? Array.isArray(currentValue)
+                                    ? currentValue
+                                    : [currentValue]
+                                : []),
+                            ...(newValue
+                                ? Array.isArray(newValue)
+                                    ? newValue
+                                    : [newValue]
+                                : []),
+                        ])
                     );
+                } else if (newValue !== currentValue && newValue != null) {
+                    // If they are different, update the value
+                    frontmatter[prop] = newValue;
                 }
-            })
-        );
+            }
+        });
     } catch (error) {
         console.error("Error in processing frontmatter: ", error);
     }


src/core/functions/internal_functions/file/InternalModuleFile.ts
```diff
@@ -210,14 +213,20 @@ export class InternalModuleFile extends InternalModule {
...
+
             try {
                 const parsed_content =
                     await this.plugin.templater.parser.parse_commands(
                         inc_file_content,
                         this.plugin.templater.current_functions_object
                     );
+                const front_matter_info = getFrontMatterInfo(parsed_content);
+                const frontmatter = yaml.load(front_matter_info.frontmatter);
+                await merge_front_matter(this.plugin.app, active_file, frontmatter);
```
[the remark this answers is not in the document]
AndyEveritt: the frontmatter is checked in `merge_front_matter` so this is not an issue

Zachatoo:
I have not forgotten this PR, I am very interested in this. I can't promise anything, but I'm hopeful that I'll have time in the next week or so to review and test this.

mdbraber:
This PR is great and exactly what I was looking for, thanks for creating this @AndyEveritt !

src/core/functions/internal_functions/file/InternalModuleFile.ts
```diff
@@ -210,14 +213,20 @@ export class InternalModuleFile extends InternalModule {
                 }
             }
 
+            const active_file = this.plugin.app.workspace.getActiveFile();
```
Zachatoo, line 216: It may not always be the active file, since you can use `tp.file.create_new` to create a new file without opening it.
I think we'll need to find a different solution for how to get `tp.file.include` to handle merging frontmatter.

src/core/Templater.ts
```diff
@@ -22,6 +24,7 @@ import {
 import { errorWrapper, errorWrapperSync, TemplaterError } from "utils/Error";
 import { Parser } from "./parser/Parser";
 import { log_error } from "utils/Log";
+import * as yaml from "js-yaml";
```
Zachatoo, line 27: We can use `parseYaml` and `stringifyYaml` from `'obsidian'` instead.

src/core/Templater.ts
```diff
@@ -302,7 +315,30 @@ export class Templater {
...
+            // Bases can create frontmatter, merge this into the template frontmatter
+            await merge_front_matter(this.plugin.app, file, frontmatter);
+            await this.plugin.app.vault.append(
+                file,
+                output_content.slice(front_matter_info.contentStart)
+            );
+            output_content = await this.plugin.app.vault.read(file);
+        } else {
+            await this.plugin.app.vault.modify(file, output_content);
+        }
```
Zachatoo, line 341: There's a lot of reading and writing going on that could potentially cause issues with other plugins. Ideally we should be doing a single atomic operation.

src/utils/Utils.ts
```diff
@@ -212,17 +226,64 @@ export function get_fn_params(func: (...args: unknown[]) => unknown) {
...
+            for (const prop in properties) {
+                const currentValue = frontmatter[prop];
+                const newValue = properties[prop];
+
+                if (currentValue === undefined) {
+                    // If the property doesn't exist, add it
+                    frontmatter[prop] = newValue;
+                } else if (
+                    Array.isArray(currentValue) ||
+                    Array.isArray(newValue)
```
Zachatoo, line 263: I'm not sure I'm sold on the idea of merging array and non-arrays together. Looking at the minified source code for Obsidian, they don't do it this way. We should consider following how core Templates works.
I'm open to differing how Templater and Templates works in this regard if it'll provide enough value, maybe do it Obsidian's way initially and change it if we get feedback that it should change?

Zachatoo (changes requested):
I've made some comments, I've also taken your work and made changes and made a PR into your branch, let me know your thoughts.
https://github.com/AndyEveritt/Templater/pull/1

Thank you so, so much for the work you've put into this. I can tell you've put a lot of thought and effort into making this feature a reality. I am very excited for this, I've started using Bases more and being able to just click "New" will make it a lot easier to manage my notes.

AndyEveritt:
I'm looking at your suggested changes now. It seems like some functionality is broken now. Will try to find the extent and cause

src/utils/Utils.ts
AndyEveritt: This was to allow it to work with single element arrays created by a base (which I expect is a bug but it is the current functionality). Currently if you have a base like this

```base
views:
  - type: table
    name: Table
    filters:
      and:
        - tags.contains("coding")

```

And you click new note, the base will create a note with this contents
```md
---
tags: coding
---
```

Instead of 
```md
---
tags:
  - coding
---
```

If you then merge a template that has an array of `tags` then you obviously want to merge them not replace the single tag created by the base with the array in the template.

The same can also exist in reverse with a base that creates a multi value array of tags and a template with only a single element.

My opinion is it is ok to deviate from the core Obsidian logic here since there is a clear bug in the obsidian template/base interaction that I expect will be fixed when Obsidian natively supports base templates

AndyEveritt: This applies to any frontmatter list property not just `tags`

Zachatoo: That is helpful context. I am fine with this deviation. Thank you for the detailed explanation!

mdbraber:
@AndyEveritt seems here that the functionality you’ve implemented at least works here, what doesn’t seem to be working? Looking forward to getting this merged in the main release!

AndyEveritt:
There seem to be a couple issues.

If you have 
`root template`
```md 
---
some_frontmatter:
---
<%- tp.file.include("[[sub template]]") %>
```

`sub template`
```md
---
sub:
---
```

That is now an error as your code needs a new line between the end of the frontmatter and the content

I think it is because you removed all the logic that handled nested template files with frontmatters using `tp.file.include`, so now in `get_frontmatter_and_content` you get content=
```
---
some_frontmatter:
------
sub:
---
```
because it is including all the sub templates then trying to merge the resulting template with the existing file.

What it should do is try to merge each sub template with the template that included it and once the root template has been fully processed, it should then be merged with the file.

I will open a new PR to fix this issue using 2.15.2 code as a base

AndyEveritt:
This is fixed in PR #1643

PR #1608, opened 2025-07-18
feat: support object-style module.exports for user scripts

### LINKED ISSUE #1607: Support object-style module.exports for multiple user functions in a single user script file
Currently, Templater only supports user scripts that export a single function via module.exports = function () {} or exports.default = function () {}.
This makes it difficult to organize related utility functions within a single file, forcing users to split their logic across multiple files unnecessarily.
While it's technically possible to export a single wrapper function that returns an object of inner functions,
this approach has several drawbacks — it creates additional overhead when calling functions (e.g., instantiating objects each time),
reduces code readability and maintainability, and complicates function usage within templates.
Moreover, this pattern is less compatible with static analysis and autocomplete tools, making development more error-prone.

I’d like Templater to support exporting an object of functions, such as:

```ts
module.exports = {
  fromNow: () => { /* ... */ },
  fromTitle: (tp) => { /* ... */ },
};
```
So it can be used in templates like this:

```ts
<%* const d = await tp.user.dateUtil.fromNow(); %>
```
This would make user scripts more modular, readable, and maintainable.

**Describe alternatives you’ve considered**
Splitting each function into its own file: this quickly becomes unmanageable as the number of functions grows.

Wrapping multiple functions inside one exported function that returns an object: this is a clumsy workaround to simulate multiple exports.

**Additional context**
Internally, this could be supported by checking if the exported value is a function, or an object where all values are functions.

For example:

```ts
typeof exported === "function"
// or
typeof exported === "object" && Object.values(exported).every(v => typeof v === "function")
```
This change would maintain backward compatibility while providing more flexibility for power users.

For reference, I’ve already implemented this functionality locally and plan to submit a pull request based on the proposed change.
I'm happy to share the implementation or request feedback ahead of the PR if needed. Thank you.

pica-git0:
# feat: support object-style `module.exports` for user scripts

### Summary

This PR adds support for exporting multiple user-defined functions from a single script file using the pattern:

```js
module.exports = {
  fn1: () => "...",
  fn2: (tp) => "..."
};
```
Previously, Templater only supported a single default-exported function per script file, which limited how utility logic could be grouped.
This change enables better organization, maintainability, and flexibility for advanced Templater users.

#### What’s Changed
- Modified UserScriptFunctions.load_user_script_function() to support both:
  - a single exported function (as before)
  - or an object of multiple functions, each of which can be used as tp.user.scriptName.functionName()
- Type-checked to ensure all values exported are functions
- Preserves backward compatibility with existing single-function user scripts

#### Motivation
Closes #1607 

Currently, grouping utility functions in one file is difficult due to the single-function export limitation.
While wrapping functions inside one function and returning them is possible, it reduces readability, hurts autocomplete support, and increases complexity unnecessarily.

Supporting object-style exports makes user scripts:

- Easier to manage
- More declarative
- Cleaner to use inside templates

It also aligns more closely with how modules are commonly structured in JavaScript.

##### Notes
- This change does not affect existing scripts using module.exports = function (...) {}.
- If module.exports is an object, all values must be functions.
- If not, an error message will inform the user that only function exports are allowed.

### DIFF
--- a/src/core/functions/user_functions/UserScriptFunctions.ts
+++ b/src/core/functions/user_functions/UserScriptFunctions.ts
@@ -61,22 +61,41 @@ export class UserScriptFunctions implements IGenerateObject {
                 err.message
             );
         }
-        const user_function = exp["default"] || mod.exports;
+        const exported = exp["default"] ?? mod.exports;
 
-        if (!user_function) {
+        if (!exported) {
             throw new TemplaterError(
                 `Failed to load user script at "${file.path}". No exports detected.`
             );
         }
-        if (!(user_function instanceof Function)) {
+
+        if (typeof exported === "function") {
+            // ✅ Case 1: The export is a single function
+            user_script_functions.set(file.basename, exported);
+        }
+        else if (typeof exported === "object" && exported !== null) {
+            // ✅ Case 2: The export is an object, check if all values are functions
+            const allValuesAreFunctions = Object.values(exported).every(
+                (v) => typeof v === "function"
+            );
+
+            if (!allValuesAreFunctions) {
+                // ❌ Error: The exported object contains non-function values
+                throw new TemplaterError(
+                    `Exported object in "${file.path}" must contain only functions.`
+                );
+            }
+
+            // ✅ Case 2-1: The export is an object and all values are functions
+            user_script_functions.set(file.basename, exported);
+        }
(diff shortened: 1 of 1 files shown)


Zachatoo (approved):
This is great! Thank you!

PR #1611, opened 2025-07-19
Fix tp.file.create_new() does not return to original context

### LINKED ISSUE #1554: tp.file.create_new() does not return to the original templater instance after execution
## Plugin information (please complete the following information):

-   OS: Windows 10]
-   Templater version: [2.10.0]
-   Obsidian version: [1.8.9]
-   Templater settings: [By default after installation]

## Describe the bug
It seems that templater does not reurn to its original instance when using tp.file.create_new, e.g:

### Simple scenario
Original.md:
```javascript
<%*
//await tp.file.create_new("MyFileContent", "MyFilename", false, tp.file.folder(true)) 

console.log(`I'm on Original: ${tp.file.path(true)}`)
%>
```
Console ouput `I'm on Original: Test/Bug_Test/Original.md`

So far ok, however if we uncomment tp.file.create_new and run it again:
```javascript
<%*
await tp.file.create_new("MyFileContent", "MyFilename", false, tp.file.folder(true)) 

console.log(`I'm on Original: ${tp.file.path(true)}`)
%>
```
Console output: `I'm on Original: Test/Bug_Test/MyFilename.md
`

### More Complex Scenario
Template running:

```javascript
tp.custom = new Object({});
tp.custom.new = "test"
let exercise_content = await tp.file.include(tp.file.find_tfile("Test/Target"));
console.log(exercise_content)
await tp.file.create_new("Test", "My_new_file", false) 
```

The Target template:
```javascript
<%* 
console.log("I reach here")
console.log(tp.custom.new)
%>
```

When running ‘Template running’ file.include returns `1:I reach here` `2:test` as it should, however if the tp.file.create_new is defined before tp.file.include it seems to ‘delete’ or lose access to the custom object even if it is defined later. For example, in this case:

Template running but with create_new defined at the beginning:
```javascript
await tp.file.create_new("Test", "My_new_file", false) 

tp.custom = new Object({});
tp.custom.new = "test"
let exercise_content = await tp.file.include(tp.file.find_tfile("Test/Target"));
console.log(exercise_content)
```

Execution output:
```
I reach here
plugin:templater-obsidian:6 Templater Error: Template parsing error, aborting. 
 Cannot read properties of undefined (reading 'new')
```

## Expected behavior
It would seem logical to me that templater returns to the original instance of templater after finishing the execution of tp.file.create_new, since it is going to continue working on the original template.

## Workaround
At this point the solution I have found is to alter the script logic to make use of tp.file.create_new() as far down the template as possible.

**Awesome plugin, Thank you!**

4Source:
I added functionality to create a backup of the current config before switching to a new one when creating a new file with ``create_new_note_from_template`` and reload it after the new file is created. 

This fixes #1554 

I tested it with an extended example from the bug
```
<%*
console.log('This message was before')
await tp.file.create_new("MyFileContent", "MyFilename", false, tp.file.folder(true))

console.log(`I'm on Original: ${tp.file.path(true)}`)
%>
```
File get's created and console prints as expected.

With current master: 
image

With my changes:
image

### DIFF
--- a/src/core/Templater.ts
+++ b/src/core/Templater.ts
@@ -74,6 +74,10 @@ export class Templater {
         };
     }
 
+    get_current_running_config(): RunningConfig {
+        return this.functions_generator.get_config();
+    }
+
     async read_and_parse_template(config: RunningConfig): Promise<string> {
         const template_content = await this.plugin.app.vault.read(
             config.template_file as TFile
@@ -164,6 +168,8 @@ export class Templater {
             return;
         }
 
+        const backup_config = this.get_current_running_config();
+
         const { path } = created_note;
         this.start_templater_task(path);
         let running_config: RunningConfig;
@@ -223,6 +229,10 @@ export class Templater {
             });
         }
 
+        await this.functions_generator.generate_object(
+            backup_config,
+            FunctionsMode.USER_INTERNAL
+        );
         await this.end_templater_task(path);
         return created_note;
     }
--- a/src/core/functions/FunctionsGenerator.ts
+++ b/src/core/functions/FunctionsGenerator.ts
@@ -19,6 +19,10 @@ export class FunctionsGenerator implements IGenerateObject {
         this.user_functions = new UserFunctions(this.plugin);
     }
 
+    get_config(): RunningConfig {
+        return this.internal_functions.get_config();
+    }
+
     async init(): Promise<void> {
         await this.internal_functions.init();
     }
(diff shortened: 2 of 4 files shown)


Zachatoo (approved):
Also tested to make sure that `tp.file.create_new` nested in a `tp.file.include` preserves the correct context and it does. Looks good to me! Thank you for your work on this!

PR #1630, opened 2025-08-28
fix: refactor to processAsync

### LINKED ISSUE #1629: Templater `app.vault.modify()` might cause race condition.

-   OS: Windows 11
-   Templater version: 2.14.1
-   Obsidian version: 1.9.12
-   Templater settings: `Trigger Templater on new file creation`

**Describe the bug**

When you have some other file creation event handler that modifies the note, `Templater` can cause race condition.

1. Files is created with content A.
2. `Trigger Templater on new file creation` setting caused `Templater` to read content `A`.
3. Another handler works and changed the content to `B`.
4. `Templater` calls `await app.vault.modify(file, 'A');` and the change B is lost.

**Expected behavior**

`Templater` has to modify note transactionally.

**Additional context**

[`process()`](https://docs.obsidian.md/Reference/TypeScript+API/Vault/process) can't be used, because `Templater` needs `async fn`, but `process()` can take only sync `fn`.

See `process()` version that supports `async fn`: https://github.com/mnaoumov/obsidian-dev-utils/blob/df6152145da3520956e2c81aa46d4a9a08c512f6/src/obsidian/Vault.ts#L304

mnaoumov:
Fixes #1629

### DIFF
--- a/src/core/Templater.ts
+++ b/src/core/Templater.ts
@@ -12,6 +12,7 @@ import {
     generate_dynamic_command_regex,
     get_active_file,
     get_folder_path_from_file_path,
+    processAsync,
     resolve_tfile,
 } from "utils/Utils";
 import TemplaterPlugin from "main";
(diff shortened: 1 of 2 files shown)


Zachatoo:
Thanks, I'm going to give the latest release some time to breathe and then I'll check this.

PR #1643, opened 2025-09-24
feat: support merging frontmatter with `tp.file.include`

AndyEveritt:
This adds support for merging template frontmatter when using `tp.file.include`

eg:

`Root`
```md
---
some_frontmatter:
root_func: <% tp.date.now() %>
not_overwritten: 1
overwritten: 1
---
<%- tp.file.include("[[Sub]]") %>

top content
```

`Sub`
```md
---
sub: 2
not_overwritten:
overwritten: 2
---

sub content
<% tp.file.include("[[Sub 2]]") %>
```

`Sub 2`
```md
---
sub_func: <% tp.date.now() %>
---
sub 2 content
```

The expected output for this template is 

```md
---
some_frontmatter:
root_func: 2025-09-24
not_overwritten: 1
overwritten: 2
sub_func: 2025-09-24
sub: 2
---
sub content
sub 2 content

top content
```

This also works with `tp.file.create_new()` and files created with Bases

Issue #1695, opened 2026-01-29
2.18.0 draws 'Cannot read properties of undefined (reading 'selection')' error in Typescript


zanodor:
**Plugin information (please complete the following information):**

-   OS: Linux, Windows 11
-   Templater version: 2.18.0
-   Obsidian version: 1.11.5

**Describe the bug**
Cannot read properties of undefined (reading 'selection') in .ts standalone plugin with Templater API dependency.
Using
```
interface TemplaterPlugin {
    templater: {
        current_functions_object: {
            file: {
                selection(): string;
            };
            system: {
                clipboard(): Promise<string>;
                prompt(prompt_text?: string, default_value?: string, throw_on_cancel?: boolean): Promise<string>;
            };
        };
    };
}
```

**Expected behavior**
Have it work as before in 2.17.1

**Additional context**
It is about no selection in markdown file and firing with URL in clipboard:
```
import * as obsidian from 'obsidian';

interface TemplaterPlugin {
    templater: {
        current_functions_object: {
            file: {
                selection(): string;
            };
            system: {
                clipboard(): Promise<string>;
                prompt(prompt_text?: string, default_value?: string, throw_on_cancel?: boolean): Promise<string>;
            };
        };
    };
}

const pasteMyLink = async (app: obsidian.App): Promise<void> => {
    const templaterPlugin = app.plugins.plugins['templater-obsidian'] as TemplaterPlugin;
    const tp = templaterPlugin?.templater?.current_functions_object;
    if (!tp) {
        new obsidian.Notice('Templater plugin not found');
        return;
    }

    const editor = app.workspace.activeLeaf?.view?.editor;
    if (!editor) return;

    const isValidUrl = async (urlString: string): Promise<boolean> => {
        try {
            new URL(urlString);
            return true;
        } catch {
            return false;
        }
    };

    const isObsidianReference = async (text: string): Promise<boolean> => {
        return /^\!?\[\[.+?(?:#[^\]]+?)?\]\]$/.test(text);
    };

    const isMarkdownLink = async (text: string): Promise<boolean> => {
        return /^\[.+?\]\(.+?\)$/.test(text);
    };

    const extractMarkdownLink = async (text: string): Promise<{ title: string; url: string } | null> => {
        const match = text.match(/^\[(.+?)\]\((.+?)\)$/);
        if (match) {
            return {
                title: match[1],
                url: match[2]
            };
        }
        return null;
    };

[...]

PR #1698, opened 2026-01-29
feat: support merging frontmatter with `tp.file.include`

AndyEveritt:
Same as #1643 but changed the backwards compatibility of `current_functions_object` to be identical to previous behaviour


Fred-Vatin:
Is there any obstacles preventing this PR to be finally merged ?

AndyEveritt:
I'm not sure why it isn't being merged tbh. I've made a release so you can just install that in your Obsidian repo if you want to use it.
https://github.com/AndyEveritt/Templater/releases/tag/2.18.1-1698

It's what I've been doing since September and it has been working fine for me on desktop and mobile

Zachatoo:
I did a test with the example from #1695 and it still isn't working with your changes, I get an empty object instead of a populated functions object and it fails.

PR #1701, opened 2026-02-27
Handle unset templates directory

sec-ml:
If `Template folder location` is not set, path names of scripts are displayed incorrectly. This only affects the display, they still run correctly.

image

Added a check for `folderLength === 0`.

I've tested locally and it works. No lint errors introduced.


Zachatoo (approved):
Looks like it's working for me, thank you!

PR #1706, opened 2026-03-13
Improve documentation for date and file modules

Yodo9000:
Make function descriptions for the date module more specific.

- change references to the "format reference" to the "Momentjs format reference"
- specify that the default value for reference dates is now/today

Just some things I found confusing when reading the documentation for the first time. Imo it's also confusing that `tp.date.now` is not called `tp.date.datetime` or similar, since (if I understood the documentation and examples correctly) 'now' is just one of its default values in practice.

### DIFF
--- a/docs/documentation.toml
+++ b/docs/documentation.toml
@@ -43,7 +43,7 @@ description = "This module contains every internal function related to dates."
 [tp.date.functions.now]
 name = "now"
 description = "Retrieves the date."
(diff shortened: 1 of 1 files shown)


Yodo9000:
Added another commit for the file module:
- Mention that the default value of filename in `tp.file.create_new` is "Untitled" (especially since it doesn't depend on the locale)
- Replace integral bars with normal vertical bars
- Clarify which file `tp.file.creation_date` refers to.

I left the "format reference[s]" as-is this time.

Edit: decided to just replace the remaining integral bars with normal ones.

Zachatoo (approved):
Looks great, thank you!

PR #1713, opened 2026-04-01
feat: add company_templates_folder setting

C-Bjorn:
## Summary

Adds a new optional `company_templates_folder` setting that allows organisations (or any user running a shared/managed vault) to maintain a second template folder whose contents take **priority** over the personal `templates_folder`.

When the setting is empty (default) Templater behaves exactly as before — fully opt-in, zero behaviour change for existing users.

---

### `src/settings/Settings.ts`
- `company_templates_folder: string` added to the `Settings` interface and `DEFAULT_SETTINGS` (default `""`)
- New `add_company_template_folder_setting()` UI method — single folder input with `FolderSuggest` autocomplete, displayed immediately below the existing "Template folder location" setting
  - Label: **"Company templates folder (MegaMem)"**
  - Helper: *"Templates here take priority over your personal templates folder."*

### `src/core/Templater.ts`
- New instance method `resolve_template_tfile(template_str: string): TFile`
  - When `company_templates_folder` is set, extracts the basename of the requested template and looks for a matching file in the company folder first (handles both `Name.md` and bare `Name` references)
  - Falls back to the existing `resolve_tfile()` behaviour when the company folder is empty or no match is found
- `on_file_creation()` updated to call `templater.resolve_template_tfile()` instead of bare `resolve_tfile()` for both **folder template** and **file regex template** resolution — ensures the company folder wins on name collision at runtime
- **Auto-exclusion guard added** — `company_templates_folder` is now automatically skipped in `on_file_creation()`, exactly as `templates_folder` is. Files created inside the company templates folder will never trigger template processing (no need to also add the folder to the "Ignore folders on file creation" list)

---

## Behaviour

| Scenario | Result |
|---|---|
| `company_templates_folder` is empty | Identical to stock Templater |
| Template name exists only in personal folder | Uses personal template (unchanged) |
| Template name exists only in company folder | Uses company template |
| Template name exists in **both** folders | **Company template wins** |
| New file created inside `company_templates_folder` | **Skipped** — same auto-exclusion as `templates_folder` |

---

## Testing
- Built clean with `pnpm run build` (334.7 kb)
- Verified in Obsidian: setting appears below "Template folder location", company templates resolve correctly, personal-only templates continue to work, empty setting leaves behaviour unchanged, files created in the company templates folder are not processed


Zachatoo:
This seems specific to your company's needs, I haven't had this requested by anyone else.

I'm going to recommend you maintain a fork of Templater for your company to include this functionality. I don't want to have to maintain this feature going forward just for you.

I'll reconsider this if other companies ask for this feature.

C-Bjorn:
Thanks for the clear feedback, Zachatoo — makes total sense. We've adopted the maintained fork model: https://github.com/C-Bjorn/Templater

This feature lives there as part of our internal tooling. Closing this PR.

PR #1717, opened 2026-04-20
fix: call super.onOpen() in SuggesterModal to restore suggestions rendering

### LINKED ISSUE #1718: 2.19.1 destroyed tp.system.suggester

-   OS: Windows 11
-   Templater version: 2.19.1
-   Obsidian version: 1.12.7
-   Templater settings: Even a fresh install on a fresh vault you can reproduce it

**Describe the bug**
If you try the example from Templater wiki:
`await tp.system.suggester(["Happy", "Sad", "Confused"], ["Happy", "Sad", "Confused"])`

on a fresh install of Templater on a fresh vault, it will return blank.

I first noticed this on my implementation of suggester:

```
<%*
let markdownFiles = tp.app.vault.getMarkdownFiles();
let pathRegex = /Engine\/Templates\/Database Entries\/[^\/]+\.md$/;

markdownFiles = markdownFiles.filter(file => {
	fileCache = tp.app.metadataCache.getFileCache(file);
	return file.path.match(pathRegex) && file.basename !== "New Database Entry"
}).toSorted();

newEntryFile = await tp.system.suggester((item) => item.basename, markdownFiles, true, "What kind of entry are you making?", 30);
%>
```

**Expected behavior**
It should display the stuff on `text_items` and `items` parameters.

**Screenshots**

Image

**Additional context**

flowing-abyss:
## Problem

Commit `063a8ac` introduced a `default_value` parameter for `tp.system.suggester` and added an `onOpen()` override to `SuggesterModal` to pre-populate the input. However, the override does not call `super.onOpen()`.

`FuzzySuggestModal` inherits `onOpen()` from `SuggestModal`, which is responsible for focusing the input element and calling `updateSuggestions()`. Without it, the modal opens but renders no suggestions — making `tp.system.suggester` completely non-functional in 2.19.1.

## Fix

Add `super.onOpen()` as the first call in the override:

```ts
onOpen(): void {
    super.onOpen(); // focuses input + populates suggestions
    if (this.default_value !== undefined) {
        this.inputEl.value = this.getItemText(this.default_value);
        this.inputEl.dispatchEvent(new InputEvent("input"));
    }
}
```

Calling `super.onOpen()` first ensures the base modal initializes correctly, then the default value is applied on top — which is also the correct order semantically.

## Impact

All users of `tp.system.suggester` are affected in 2.19.1. Downgrading to 2.19.0 is the current workaround.

### DIFF
--- a/src/core/functions/internal_functions/system/SuggesterModal.ts
+++ b/src/core/functions/internal_functions/system/SuggesterModal.ts
@@ -24,6 +24,7 @@ export class SuggesterModal<T> extends FuzzySuggestModal<T> {
     }
 
     onOpen(): void {
+        super.onOpen();
         if (this.default_value !== undefined) {
             this.inputEl.value = this.getItemText(this.default_value);
             this.inputEl.dispatchEvent(new InputEvent("input"));


Zachatoo (approved):
Thank you for fixing this!

PR #1744, opened 2026-05-30
fix: resolve #1259 — Folder Templates don't work on selected Template Folder location

### LINKED ISSUE #1259: Folder Templates don't work on selected Template Folder location

-   OS: Windows 10
-   Templater version: 1.18.1
-   Obsidian version: 1.4.16
-   Templater settings: Template folder location: "Templates/"

**Describe the bug**
If the "Templates/" folder is marked as the template folder location, then Folder Templates for the folder "Templates/" do not work. If I RClick on the "Templates/" folder and select New Note, nothing happens, and a normal Untitled blank note is created.

If I removed the "Templates/" folder from the template folder location, the intended result is achieved.

**Expected behavior**
I have a Template that generates other Templates, and that should work as intended, since I've set the folder template to be used on that folder. Despite having the template folder location set, which is useful for when I want to insert a template elsewhere, it should not stop me from being able to dynamically generate other notes in that specified folder using Templater.

andoan16:
## Summary

fix: resolve #1259 — Folder Templates don't work on selected Template Folder location

## Problem

**Severity**: `High` | **File**: `src/handlers/EventHandler.ts`

The EventHandler likely contains logic that excludes the template folder from folder template processing to prevent conflicts, but this prevents legitimate use cases where users want to generate templates within the template folder. The fix should allow folder templates to work on the template folder while still preventing infinite recursion.

## Solution

Review the event handling logic for new note creation and folder template application. Look for conditions that exclude the template folder path and modify them to allow folder templates to be applied while maintaining safeguards against infinite recursion. The logic should distinguish between inserting templates into the template folder (which should be allowed) versus creating recursive template scenarios.

## Changes

- `src/handlers/EventHandler.ts` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/src/handlers/EventHandler.ts
+++ b/src/handlers/EventHandler.ts
@@ -71,12 +71,15 @@ export default class EventHandler {
         if (this.settings.trigger_on_file_creation) {
             this.trigger_on_file_creation_event = this.plugin.app.vault.on(
                 "create",
-                (file: TAbstractFile) =>
-                    Templater.on_file_creation(
-                        this.templater,
-                        this.plugin.app,
-                        file,
-                    ),
+                (file: TAbstractFile) => {
+                    if (file instanceof TFile) {
+                        Templater.on_file_creation(
+                            this.templater,
+                            this.plugin.app,
+                            file,
+                        );
+                    }
+                },
             );
             this.plugin.registerEvent(this.trigger_on_file_creation_event);
         } else {


Zachatoo (changes requested):
Can you help me understand how this resolves the issue? This change does not do what we decided the correct solution was in the related issue.

Zachatoo:
Closing due to no response, feel free to reopen if you make changes

PR #1753, opened 2026-06-18
feat: add autocomplete for object-exported user script functions

### LINKED ISSUE #1752: Add autocomplete when exporting object from user script
**Plugin information:**

-   OS: macOS 26.5.1
-   Templater version: 2.20.5
-   Obsidian version: 1.12.7

**Describe the bug**
I have a user script file, but instead of exporting a single function, I export object with multiple functions. So for example I have a `foo.js` that looks like this
```
module.exports = {
  func: () => console.log("test")
}
```
If I type `tp.user.foo` there is no autocomplete suggesting `.func`

**Expected behavior**
Autocomplete suggests my exported functions

**Screenshots**

**Additional context**

jianongHe:
Fixes #1752.

## Summary

This adds autocomplete for user scripts that export an object. Previously, `tp.user.` could suggest the script file, but `tp.user.<script>.` did not suggest function members from an object export. With this change, Templater loads the selected user script and suggests function-valued object members such as `func` and `doSomething`.

The PR also surfaces JSDoc/TSDoc-style documentation for documented object members, including parameter and return metadata in the completion details.

## Manual verification script

I used this minimal user script for the UI verification screenshots:

```js
/**
 * Does something from an object export.
 * @param name - The name to use
 * @returns The result message
 */
function doSomething(name) {
    return "Something was done " + name;
}

module.exports = {
    func: () => console.log("test"),
    doSomething,
};
```

## UI verification

Manual UI verification was done with Templater 2.20.5 + Obsidian 1.12.7, matching the version reported in the issue. The current development build requires Obsidian 1.13.0, which I could not download locally without an Obsidian Insiders account, so I ported this fix into a temporary 2.20.5 checkout for the manual UI check.

<table>
  <tr>
    <td align="center"><strong>Before: script file is suggested</strong></td>
    <td align="center"><strong>Before: object members are missing</strong></td>
  </tr>
  <tr>
    <td>Before: tp.user.fo suggests the foo script</td>
    <td>Before: tp.user.foo. does not suggest object members</td>
  </tr>
</table>

<strong>After: object members and documentation are suggested</strong>

After: tp.user.foo. suggests func and doSomething with documentation

## Verification

- `pnpm typecheck`
- `pnpm lint`
- `git diff --check`
- `pnpm build`
- Manual UI verification with Templater 2.20.5 + Obsidian 1.12.7 as shown above

## Notes

This uses the existing user script loader to inspect the target script export shape for nested autocomplete, so completing `tp.user.<script>.` evaluates that user script in the same way runtime loading does.

I did not claim a passing WDIO run: a focused local WDIO run hung before producing spec output in this environment.

### DIFF
--- a/src/editor/TpDocumentation.ts
+++ b/src/editor/TpDocumentation.ts
@@ -1,6 +1,14 @@
 import TemplaterPlugin from "main";
+import { TFile } from "obsidian";
 import { errorWrapper } from "utils/Error";
-import { get_fn_params, get_tfiles_from_folder, is_object, populate_docs_from_user_scripts } from "utils/Utils";
+import { UserScriptFunctions } from "core/functions/user_functions/UserScriptFunctions";
+import type { UserScriptFunction } from "types";
+import {
+    get_fn_params,
+    get_tfiles_from_folder,
+    is_object,
+    populate_docs_from_user_scripts,
+} from "utils/Utils";
 import documentation from "../../docs/documentation.toml";
 
 const module_names = [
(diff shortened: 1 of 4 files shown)


Zachatoo (approved):
Awesome, all tests passing and manual tests look good, thanks for adding the test even though you couldn't test it yourself!

Image

PR #1758, opened 2026-07-08
Add configurable maxBuffer for user system commands

### LINKED ISSUE #1749: Hitting "stdout maxBuffer length exceeded" on system commands

-   Windows 11
-   Templater version: 2.22.1
-   Obsidian version: 1.12.7
-   Templater settings: `tp.user.<user_defined_system_command>()`

**Describe the bug**
I'm not sure if I would classify this as a bug or feature request, but the issue I'm experiencing is when I run a [user-defined system command](https://silentvoid13.github.io/Templater/user-functions/system-user-functions.html) which outputs a significant amount of text (my function outputs 1.2mb of text) you get the error:

```
plugin:templater-obsidian:6 Templater Error: Error with User Template 
 stdout maxBuffer length exceeded
```

I believe this is because when we use `child_process` in [UserSystemFunctions.ts#L28](https://github.com/SilentVoid13/Templater/blob/master/src/core/functions/user_functions/UserSystemFunctions.ts#L28) `exec`, it uses the default `maxBuffer`, which is 1 million bytes (see [docs](https://nodejs.org/api/child_process.html#child_processexeccommand-options-callback)). I wonder if we can add an option to specify `maxBuffer` or we can avoid this error entirely (I think, I would need to test this) with [child_process.spawn](https://nodejs.org/api/child_process.html#child_processspawncommand-args-options) as that uses a stream instead of storing the output in a buffer.

**Expected behavior**
- Perhaps we can add an option to templater to manually specify the `maxBuffer`
- Another option (and perhaps the better one) is using `child_process.spawn` instead of `child_process.exec`, which uses stream and doesn't wait for the entire output to buffer.

CornElDos:
## Description

User system commands (User System Functions) run through Node's `child_process.exec`, which caps combined stdout at a default `maxBuffer` of 1 MiB and rejects with `stdout maxBuffer length exceeded` once a command produces more output than that. This is the failure reported in #1749.

This adds a `command_max_buffer` setting (default 100 MiB, in bytes) and passes it as `maxBuffer` in the exec options, so commands with large output no longer fail while the limit stays tunable. It mirrors the existing advanced `command_timeout` / `shell_path` settings (data-model settings without dedicated UI), so no migration is needed — `Object.assign({}, DEFAULT_SETTINGS, raw)` fills the default in for existing users.

Fixes #1749

## Testing

- `pnpm run typecheck` — pass
- `pnpm run lint` — pass
- `pnpm run build` — pass

I was not able to run the e2e suite locally, as it requires Obsidian 1.13 (Catalyst) per the discussion in the issue thread. Happy to adjust the default buffer size or the overall approach based on your preference.

### DIFF
--- a/src/core/functions/user_functions/UserSystemFunctions.ts
+++ b/src/core/functions/user_functions/UserSystemFunctions.ts
@@ -82,6 +82,7 @@ export class UserSystemFunctions implements IGenerateObject {
                                 this.plugin.settings.command_timeout * 1000,
                             cwd: this.cwd,
                             env: process_env,
+                            maxBuffer: this.plugin.settings.command_max_buffer,
                             ...(this.plugin.settings.shell_path && {
                                 shell: this.plugin.settings.shell_path,
                             }),
--- a/src/settings/Settings.ts
+++ b/src/settings/Settings.ts
@@ -42,6 +42,10 @@ export interface IgnoreFolderOnCreation {
 export const DEFAULT_SETTINGS: Settings = {
     data_version: 2,
     command_timeout: 5,
+    // Maximum stdout/stderr size (bytes) for user system commands.
+    // Overrides Node's 1 MiB `exec` default, which made commands with large
+    // output fail with "stdout maxBuffer length exceeded".
+    command_max_buffer: 1024 * 1024 * 100,
     templates_folder: "",
     templates_pairs: [],
     trigger_on_file_creation_mode: "none",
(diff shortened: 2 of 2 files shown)


Zachatoo (changes requested):
I'm good with this approach, and I'm okay that you didn't add tests since you don't have access to the beta versions of Obsidian.

It doesn't look like you added UI to manage the setting, is that intentional? I feel like we'd want this added in settings. Somewhere instead of just hidden in the data.json file.

Image

PR #1761, opened 2026-07-10
feat(prompt): Added ability to auto select default text in system prompt

### LINKED ISSUE #1760: Add an optional system prompt argument to auto select the defaulted text value
When using `tp.system.prompt`, you are able to optionally add a default value.

It would be a nice quality of life feature to add an argument so that default text that is displayed in the prompt can optionally be automatically selected so allow for easy overwriting.

 I have a workflow now that I have a prompt that has "Please add details here..." which is meant to be overwritten. It would be nice for it to be auto selected so I can just begin typing.

srz2:
# feat(system): add `select_default_value` option to `tp.system.prompt`

## Summary

Adds a new optional argument, `select_default_value`, to `tp.system.prompt`. When
set to `true`, the default value shown when the prompt modal opens is
selected/highlighted, so the user can immediately overwrite it by typing. When
`false` (the default), behavior is unchanged.

**New signature:**

```js
tp.system.prompt(
  prompt_text?,
  default_value?,
  throw_on_cancel = false,
  multiline = false,
  select_default_value = false
)
```

## Motivation

When a prompt is opened with a default value, users often want to replace it
rather than append to it. Without this option, the cursor is placed at the end
of the pre-filled text, forcing the user to manually clear it first.

## Usage

```js
// Opens with "Input a mood..." pre-selected so typing replaces it
example = """<% await tp.system.prompt("What is your mood today?", "Input a mood...", false, false, true) %>"""
```

## Backward compatibility

Fully backward compatible. The new parameter is optional and defaults to
`false`, so existing calls to `tp.system.prompt` are unaffected.

Closes #1760

### DIFF
--- a/docs/documentation.toml
+++ b/docs/documentation.toml
@@ -514,7 +514,7 @@ example = "<% tp.system.clipboard() %>"
 [tp.system.functions.prompt]
 name = "prompt"
 description = "Spawns a prompt modal and returns the user's input."
-definition = "tp.system.prompt(prompt_text?: string, default_value?: string, throw_on_cancel: boolean = false, multiline?: boolean = false)"
+definition = "tp.system.prompt(prompt_text?: string, default_value?: string, throw_on_cancel: boolean = false, multiline?: boolean = false, select_default_value?: boolean = false)"
 
 [[tp.system.functions.prompt.args]]
 name = "prompt_text"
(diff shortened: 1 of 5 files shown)


Zachatoo (approved):
Looks great, and tests pass. Thanks!

